PR #1, opened 2026-06-23
Transition solvers extraction to CUDA-Q Algorithms

wsttiger:
## Summary

This PR introduces the extracted CUDA-Q Algorithms library layout on a PR-compatible branch derived from NVIDIA main.

It includes:
- standalone CUDA-Q Algorithms CMake/package structure
- fermion mappings, including Jordan-Wigner and Bravyi-Kitaev support
- state-preparation primitives and operator-pool helpers
- C++ and Python bindings/tests for the extracted algorithms surface
- cleaned-up scripts for the algorithms-only repository layout

## Notes for reviewers

The source transition branch was wsttiger:transition/solvers-extraction. That branch did not share history with NVIDIA:main, so this PR branch uses NVIDIA main as its parent while matching the transition branch file tree.

The latest cleanup removed legacy CUDA-QX/Solvers/QEC wheel, CI, container, and documentation scripts that no longer matched the extracted algorithms repository.

## Validation

- bash -n scripts/test_algorithms_build.sh scripts/run_clang_format.sh scripts/run_yapf_format.sh
- bash scripts/run_clang_format.sh
- bash scripts/run_yapf_format.sh
- bash scripts/test_algorithms_build.sh --build-dir /tmp/cudaq_algorithms_script_check --cudaq-prefix /usr/local/cudaq
- CTest: 27/27 passed
- Python tests: 21 passed


ivanbasov:
The README notes that VQE, ADAPT-VQE, QAOA, GQE, PySCF drivers, and optimizers are intentionally excluded from this initial extraction. Can you confirm the longer-term intent: are these planned to be ported into this repo in a later PR, or are they being dropped entirely from the Algorithms line? This determines whether reviewers should treat the omissions here as "not yet" vs. "out of scope permanently," and whether we should preserve any of that code/history before it is pruned.

ivanbasov:
For a public repo this is missing the top-level governance files that the source repo carried — `LICENSE`, `NOTICE`, `CONTRIBUTING`, the CLA, and `CITATION.cff` — and this PR also removes the template `SECURITY.md`. These should be added (or restored) before the repo goes public. Flagging in case it is already tracked separately.

ivanbasov:
A few CI observations (understood it is still in progress):
- The repo ships both `pyproject.toml.cu12` and `pyproject.toml.cu13`, but the lib/wheel workflows only matrix on `cuda_version: ['12.6']`. The cu13 packaging path is currently unbuilt/untested — add a cu13 leg or confirm it is intentionally deferred.
- Action versions are inconsistent: `ci.yaml` uses `actions/checkout@v6` / `setup-python@v6` while `lib_algorithms.yaml` / `wheel_algorithms.yaml` use `@v4`. Pin to one version.
- Please confirm the stateprep `__qpu__` kernels are actually executed under test (CPU simulator), not just compiled — that is where any namespace/mangling issue would surface.

ivanbasov:
Comparing against the source library, several tests covering functionality that *was* carried over appear dropped (beyond the expected removal of NISQ/registry tests):
- `test_bravyi_kitaev`: the Seeley-Richard-Love coverage went from 11 cases to 5 in the consolidated `SeeleyRichardLoveCases`. Were the other 6 cases dropped intentionally?
- `test_uccgsd_operator_pool`: lost `RegressionTestForOrderingBug`, `OperatorsAreAntiHermitian`, `Singles`/`DoublesOnlyGeneration`, `AllUniquePairingsGenerated`, and the singles/doubles-count check.
- `test_ceo_operator_pool` / `test_upccgsd_operator_pool`: lost the `OperatorsAre(Anti)HermitianGenerators` physics-property checks.
- Python stateprep coverage dropped from ~19 test functions (ceo/uccsd/uccgsd/upccgsd) to 9 in `test_stateprep.py`.

The Hermiticity/anti-Hermiticity property checks and the ordering-bug regression test are worth keeping even with the new free-function API. Could these be restored or ported?

include/cudaq/algorithms/stateprep/uccsd.h
```diff
@@ -0,0 +1,89 @@
...
+ * This source code and the accompanying materials are made available under    *
+ * the terms of the Apache License 2.0 which accompanies this distribution.    *
+ ******************************************************************************/
+#pragma once
+
+#include "cudaq.h"
+#include <tuple>
+#include <vector>
+
+namespace cudaq_algorithms::stateprep {
```
ivanbasov, line 24: This file declares the kernels/types in a flat `cudaq_algorithms::stateprep` namespace and then re-exports them into `cudaq::algorithms::stateprep` via `using` declarations. The three sibling device kernels (`ceo.cpp`, `uccgsd.cpp`, `upccgsd.cpp`) declare their `__qpu__` kernels directly in `cudaq::algorithms::stateprep` and appear to compile fine. Is the flat-namespace + alias here a deliberate workaround for an nvq++ mangling issue specific to the uccsd kernels, or leftover from the rename? If it is needed, it should be applied (and commented) consistently across all four; if not, normalize uccsd to match its siblings.

wsttiger:
> The README notes that VQE, ADAPT-VQE, QAOA, GQE, PySCF drivers, and optimizers are intentionally excluded from this initial extraction. Can you confirm the longer-term intent: are these planned to be ported into this repo in a later PR, or are they being dropped entirely from the Algorithms line? This determines whether reviewers should treat the omissions here as "not yet" vs. "out of scope permanently," and whether we should preserve any of that code/history before it is pruned.

Removed permanently - this library is focused on FTQC primitives. VQE, ADAPT-VQE, QAOA are primarily for the NISQ era computing.

include/cudaq/algorithms/stateprep/uccsd.h
wsttiger: Good catch, and it turned out to be load-bearing. I first normalized `uccsd` to declare directly in `cudaq::algorithms::stateprep` like its siblings — but **that crashes the nvq++ quake bridge** (`QuakeBridgeVisitor::VisitCallExpr`, abort) when the `uccsd` kernel calls its sibling `single_excitation`/`double_excitation` kernels. The original flat `cudaq_algorithms::stateprep` namespace + `using`-re-export compiles cleanly.

So it's a real, required workaround specific to the uccsd kernels — not rename leftover. The siblings don't hit it because they don't have the inter-kernel `__qpu__` call pattern that triggers the faulty mangling path. Per your suggestion ("if it is needed, it should be applied and commented"), I've kept the structure and added a comment in `uccsd.h` documenting why, with a note to revisit once the nvq++ bug is fixed. The public API remains `cudaq::algorithms::stateprep::*`. (2128fd2)

wsttiger:
Thanks — confirming this is captured durably now. The README has been updated to state that VQE, ADAPT-VQE, QAOA, GQE, PySCF drivers, and optimizers are **permanently out of scope** (this library targets FTQC primitives), not deferred, with a pointer that they remain available in `cudaqx/libs/solvers`. (2128fd2)

wsttiger:
Added in 2128fd2:
- `LICENSE` (full Apache-2.0), `NOTICE`, `SECURITY.md` (restores the removed template), `Contributing.md`, `Contributor_License_Agreement.md`, `CITATION.cff`, and `Code_of_Conduct.md`.

Two notes for review:
- **`NOTICE`** is intentionally *not* a copy of the cudaqx NOTICE — that one attributes Stim, PyMatching, PRIMA, etc., which this library doesn't pull in. I scoped it to this repo's actual third-party deps (CUDA-Q, nanobind, GoogleTest). Please sanity-check the dependency list.
- I included `Code_of_Conduct.md` as well (standard for NVIDIA OSS repos) even though it wasn't named — happy to drop it if you'd prefer.

wsttiger:
Addressed in 2128fd2:
- **Action versions:** unified `actions/setup-python` to `@v6` (`ci.yaml` was on `@v5`); `actions/checkout@v4` was already consistent everywhere.
- **cu13:** the build scripts already handle `pyproject.toml.cu13`; only the CI matrices were 12.6-only. Rather than add an untested leg, I documented cu13 as intentionally deferred in both `lib_algorithms.yaml` and `wheel_algorithms.yaml` matrices, until a cu13 devcontainer/devdeps image is available (one-line change to enable). Can file a tracking issue if you'd like.
- **stateprep kernels executed under test:** yes — `tests/python/test_stateprep.py` runs `cudaq.sample()` on the `uccsd`/`uccgsd`/`upccgsd`/`ceo` kernels (the smoke + fixed-parameter regression tests), so they execute on the CPU simulator, not just compile.

wsttiger:
Restored in 2128fd2, adapted to the free-function API and verified to build and pass:
- **`test_bravyi_kitaev`:** Seeley-Richard-Love coverage back to all 11 cases (added (0,5), (18,19), (11,5), (7,9), (9,15), (3,7)).
- **`test_uccgsd_operator_pool`:** restored `RegressionTestForOrderingBug`, `OperatorsAreAntiHermitian`, `CorrectSinglesAndDoublesCount`, `SinglesOnlyGeneration`, `DoublesOnlyGeneration`.
- **`test_ceo_operator_pool` / `test_upccgsd_operator_pool`:** restored the `OperatorsAre(Anti)HermitianGenerators` property checks (+ singles/doubles count and doubles-only for upccgsd).
- **Python `test_stateprep.py`:** added an open-shell (spin>0) excitation-count test.

On the remaining Python delta: the rest of the dropped Python tests (`test_solvers_*`, the `adapt`/`vqe` cases, `create_molecule`) depend on VQE/ADAPT and the PySCF driver, which are out of scope per the scope discussion, so they weren't ported. The in-scope ones (the fixed-parameter loops regression and active-space excitation assertions) were already present.

PR #3, opened 2026-06-25
Add Suzuki-Trotter Hamiltonian simulation primitives

wsttiger:
This PR adds Suzuki-Trotter product-formula support for Hamiltonian time evolution using CUDA-Q’s existing exp_pauli() primitive.

The main addition is a reusable apply_trotter(...) primitive that applies first-, second-, or fourth-order Suzuki-Trotter evolution for Pauli-sum Hamiltonians. The default order is second order, with first order and fourth-order Forest-Ruth composition also supported. The implementation is intended as a building block for Hamiltonian simulation workflows rather than a standalone application-level API.

Highlights:

- Adds C++ Suzuki-Trotter implementation built around cudaq::exp_pauli().
- Supports first-, second-, and fourth-order product formulas.
- Exposes the functionality to Python through nanobind.
- Adds C++ and Python tests comparing Trotterized evolution against exact classical evolution.
- Includes validation on a nontrivial 4-qubit Hamiltonian with multiple Pauli terms.
- Documents the expected global-phase behavior for comparisons against exact evolution.

This fits into the broader CUDA-Q Algorithms direction by packaging Hamiltonian simulation as a reusable FTQC/NISQ-adjacent primitive that can support examples and higher-level workflows without baking in a specific application.


ivanbasov (commented):
Thanks for this — overall a clean, well-isolated addition, and the test design is genuinely strong (the commuting-Hamiltonian tests compare against analytic/exact diagonalization rather than a re-implemented product formula, so they catch sign/coefficient/angle errors; the non-symmetric Hamiltonian + non-symmetric initial state also implicitly guards endianness). The global-phase / `identity_coefficient` semantics are documented unusually thoroughly. The math (orders 1/2/4, Forest-Ruth weights with `2*w1 + w0 = 1`) checks out.

My comments are mostly maintainability/portability rather than correctness. Priorities before merge would be the hardcoded constants, the missing license headers, and the linker-flag portability question; the dual-namespace monkey-patching is more of a design discussion.

One minor non-inline note: `apply_trotter` silently no-ops on invalid runtime input (unsupported order, mismatched lengths, zero steps). That's reasonable for a QPU primitive that can't throw on-device and it's well documented/tested, but a direct caller passing `order=3` gets identity back with no signal. A debug-build assert could surface misuse earlier. Non-blocking.

lib/hamiltonian_simulation/device/trotter.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+  const double dt = time / static_cast<double>(steps);
+  for (std::size_t step = 0; step < steps; ++step) {
+    if (order == 1) {
+      for (std::size_t i = 0; i < words.size(); ++i)
+        exp_pauli(-dt * coefficients[i], qubits, words[i]);
+    } else if (order == 4) {
+      for (std::size_t i = 0; i < words.size(); ++i) {
+        // double forest_ruth_w1 = 1.3512071919596578;
+        exp_pauli(-0.5 * 1.3512071919596578 * dt * coefficients[i], qubits,
```
ivanbasov, line 31: The header defines `forest_ruth_w1` / `forest_ruth_w0` as `inline constexpr double`, but the order-4 path inlines the literals `1.3512071919596578` / `-1.7024143839193153` six times, each shadowed by a `// double forest_ruth_w1 = ...` comment. This reads like a workaround for an NVQ++ kernel-lowering limitation on referencing namespace-scope `constexpr` from device code.

Is that the reason? If so, let's document it explicitly at the call site. If not, please use the named constants — as written, editing one of the six copies silently desyncs from the others, and the constants in the header become dead.

lib/hamiltonian_simulation/device/trotter.cpp
```diff
@@ -0,0 +1,70 @@
...
+                  words[i]);
+      }
+      for (std::size_t i = words.size(); i > 0; --i) {
+        // double forest_ruth_w1 = 1.3512071919596578;
+        exp_pauli(-0.5 * 1.3512071919596578 * dt * coefficients[i - 1], qubits,
+                  words[i - 1]);
+      }
+    } else {
```
ivanbasov, line 61: Efficiency only, non-blocking: standard symmetric compositions merge the trailing half-rotation of one `S2` with the leading half of the next — here that's the order-4 sub-step boundaries and the multi-step boundary for order 2 (both apply the same term back-to-back, e.g. `exp(-0.5*dt*c0)` then `exp(-0.5*dt*c0)`). Not merging is fully correct, it just emits extra `exp_pauli` rotations per term. Fine to leave as a follow-up, but worth a note.

lib/unittests/CMakeLists.txt
```diff
@@ -38,8 +38,10 @@ function(cudaqa_add_unittest test_name)
       GTest::gtest_main
       cudaq-algorithms
       cudaq::cudaq
+      cudaq::cudaq-${CUDAQ_TARGET}-target
   )
   add_dependencies(CUDAQAlgorithmsUnitTests ${test_name})
+  target_link_options(${test_name} PRIVATE LINKER:--no-as-needed)
```
ivanbasov, line 44: Two concerns here:

1. **Portability**: `LINKER:--no-as-needed` is a GNU-ld/lld flag; macOS `ld` rejects `--no-as-needed`. If CI covers macOS this will fail to link.
2. **Scope**: this `target_link_options` and the `cudaq::cudaq-${CUDAQ_TARGET}-target` link on line 41 are added to the shared `cudaqa_add_unittest` function, so they apply to *every* unittest, not just `test_trotter`. If the simulator-backend link is only needed because `test_trotter` calls `get_state`, consider scoping it to that target. Can you confirm this doesn't regress the other platforms' builds?

python/bindings/py_hamiltonian_simulation.cpp
```diff
@@ -0,0 +1,86 @@
+#include "cudaq_algorithms.h"
```
ivanbasov, line 1: Missing the NVIDIA Apache-2.0 license header that every other new file in this PR carries. Same for `python/cudaq_algorithms/_hamiltonian_simulation.py` and `python/cudaq_algorithms/hamiltonian_simulation/__init__.py` — likely a repo-convention / CI requirement.

python/cudaq_algorithms/_hamiltonian_simulation.py
```diff
@@ -0,0 +1,258 @@
+import math
```
ivanbasov, line 1: Missing the NVIDIA Apache-2.0 license header (see the example/test files in this PR for the canonical block).

python/cudaq_algorithms/hamiltonian_simulation/__init__.py
```diff
@@ -0,0 +1 @@
+from cudaq_algorithms._hamiltonian_simulation import *
```
ivanbasov, line 1: Missing the NVIDIA Apache-2.0 license header.

python/cudaq_algorithms/_hamiltonian_simulation.py
```diff
@@ -0,0 +1,258 @@
...
+        num_terms=len(words),
+        steps=steps,
+        order=order,
+        pauli_rotations=rotations,
+        estimated_cx_count=estimated_cx_count,
+        identity_coefficient=float(identity_coefficient))
+
+
+_cpp_hamiltonian_simulation.make_trotter_terms = make_trotter_terms
+_cpp_hamiltonian_simulation.make_trotter_plan = make_trotter_plan
```
ivanbasov, line 241: This block attaches the pure-Python `make_trotter_terms` / `make_trotter_plan` / `estimate_trotter_resources` / dataclasses onto the C++ submodule object. The effect is that `algorithms.hamiltonian_simulation.X` (the C++ submodule from `_pycudaq_algorithms`) and `cudaq_algorithms.hamiltonian_simulation.X` (this Python subpackage) are made to look identical by side effect.

It works and the tests exercise both paths, but it's fragile — the next maintainer has to discover that two distinct `hamiltonian_simulation` namespaces are being reconciled by monkey-patching. Could the public surface be a single clean module instead (e.g. re-export the C++ `apply_trotter` from the Python subpackage and drop the patching)?

tests/python/test_trotter.py
```diff
@@ -0,0 +1,495 @@
...
+
+    for order in (1, 2, 4):
+        state = np.asarray(cudaq.get_state(evolve, coefficients, words, 0.8, 3,
+                                           order),
+                           dtype=np.complex128)
+        expected = _simulate_trotter(coefficients, words, identity, num_qubits,
+                                     0.8, 3, order, ket)
+        np.testing.assert_allclose(state, expected, atol=1e-6)
+
+
```
ivanbasov, line 326: This test runs on the *default* target and asserts `errors[4] < 5.0e-6` plus the ordering `errors[4] < errors[2]`. The `double_precision_target` fixture's own comment notes the default simulator is single precision with a ~1e-7 floor, which is exactly why the scaling and commuting tests opt into `qpp-cpu`. If the true order-4 error here sinks near that floor, the tight threshold/ordering could flake. Suggest adding the same `double_precision_target` fixture for consistency.

PR #4, opened 2026-06-27
Add Linear Combination of Unitaries (LCU), qubitization, and quantum singular value transformation (qsvt)

wsttiger:
## Summary

This PR adds foundational FTQC primitives for Pauli LCU block encodings, qubitization, and QSVT/QSP-style sequence construction.

The main goal is to make the existing QEL work reusable as lower-level algorithmic building blocks rather than a standalone application path. The PR exposes PauliLCU metadata and kernel data, reusable PREPARE/SELECT/UNPREPARE-style block-encoding helpers, qubitization walk primitives, QSVT/QSP signal-phase and sequence helpers, and Python-facing utilities for composing these primitives in examples and tests.

## Key Additions

- Adds `PauliLCU` block encoding support for `cudaq::spin_op` Hamiltonians.
- Adds reusable block-encoding kernel helpers for PREPARE, SELECT, UNPREPARE, and APPLY.
- Adds qubitization primitives, including walk, adjoint walk, walk powers, and controlled variants.
- Adds QSVT/QSP signal phase and sequence helpers.
- Adds Python APIs for composing PauliLCU, qubitization, and QSVT primitives.
- Adds QSVT Hamiltonian simulation examples using PauliLCU block encodings.
- Adds Krylov/QEL support utilities and examples using the new primitives.
- Adds tests comparing block-encoding, qubitization, QSVT, and QEL behavior against dense NumPy / exact references.

## Notes

The QSVT/QSP phase-generation problem is intentionally not solved in this PR. Instead, this PR establishes the primitive layer needed to consume externally generated phase sequences, including phases from tools such as QSPPACK.

The qubitization walk convention is documented explicitly: the device walk evaluates the response at `-H / alpha` for the PauliLCU Hamiltonian convention used here. Tests now pin this behavior so users and future development have a clear reference point.

## Validation

- Full C++ build passes.
- `ctest`: 63/63 passed.
- Python tests: 46 passed.
- Formatting checked with `run_clang_format.sh` and `run_yapf_format.sh`.


lib/block_encoding/device/kernels.cpp
```diff
@@ -0,0 +1,409 @@
...
+      if (code == 1)
+        x<cudaq::ctrl>(ancilla, system[q_idx]);
+      else if (code == 2)
+        y<cudaq::ctrl>(ancilla, system[q_idx]);
+      else if (code == 3)
+        z<cudaq::ctrl>(ancilla, system[q_idx]);
+    }
+
+    if (sign < 0)
+      CUDAQ_ALGORITHMS_APPLY_Z_BY_ARITY(ancilla, n_ancilla);
```
kvmto: **[Blocking] Sign correction silently dropped for single-term (0-ancilla) negative-coefficient decompositions.**

`CUDAQ_ALGORITHMS_APPLY_Z_BY_ARITY`'s if/else-if ladder covers `COUNT` 1–10 with **no `COUNT==0` branch**, so when the LCU retains exactly one non-identity term (`num_ancilla == 0` — a genuinely single-term Hamiltonian, or one term left after coefficient thresholding / `include_identity=false`) with a negative coefficient, this sign correction expands to nothing. With zero ancilla qubits there is no register to project out, so `apply()`'s output *is* the encoded operator: the circuit encodes `+P₀/α` where the documented contract (`pauli_lcu.h`: `⟨0|_anc U |0⟩_anc = H/α`) requires `−P₀/α`.

Verified by compiling this branch and comparing the full `get_state` matrix of `apply()` for `h = +0.5·X₀Z₁` vs `h = −0.5·X₀Z₁`: **bit-for-bit identical** when they must be negatives of each other. `controlled_select` is unaffected — `CUDAQ_ALGORITHMS_APPLY_CONTROLLED_Z_BY_ARITY` does have a `COUNT==0` branch (`z(CONTROL)`), and the same harness confirms it produces correctly negated results.

Blast radius: `pauli_lcu::select()`/`apply()` (Python-bound), and everything above them — `qubitization::apply_walk`/`apply_adjoint_walk`, `qsvt::apply_phase_sequence`, and the header-based walk. Downstream physics is silently computed for `−H` instead of `H`.

Suggested fix: handle `num_ancilla==0` + negative sign explicitly — fold the sign into the Pauli application via anticommuting conjugation, or reject at `decompose_lcu` time with a clear error. Either way, please add a regression test: single-term negative-coefficient `pauli_lcu` validated via `get_state` against the dense Pauli action (mirroring `checkBlockEncodingMatchesDenseHamiltonianAction`).

include/cudaq/algorithms/detail/qpu_dispatch.h
```diff
@@ -0,0 +1,207 @@
...
+// CONTRACT: COUNT must be in [0, max_qpu_dispatch_qubits]. These ladders are
+// `if/else if` chains with NO terminal `else`, so a COUNT outside the handled
+// range silently expands to a no-op (no gate, no error) -- because device
+// (`__qpu__`) code cannot throw. Callers are responsible for keeping COUNT in
+// range. In practice every register routed through these macros is an LCU
+// ancilla / QSVT signal register whose size is `decompose_lcu`-validated
+// against `max_lcu_ancilla_qubits` (see pauli_lcu.cpp), so the cap is enforced
+// on the host before any kernel runs. A COUNT of 0 (single-term LCU, zero
+// ancilla) is intentionally a no-op for the uncontrolled ladders: the dropped
+// sign there is only an unobservable global phase. If you add a new caller
```
kvmto: **[Blocking — companion to the kernels.cpp sign-drop comment]** This rationale is correct for `reflect_about_zero` on an empty register (a 1-dimensional space, genuinely a global phase) but **factually wrong for the SELECT sign-correction call site** (`block_encoding::select`, kernels.cpp:126-127). There, with zero ancilla qubits, the dropped `Z` changes the encoded operator itself (`+P₀` vs `−P₀`), not an unobservable phase — confirmed by compiled execution. Please scope this comment to the reflection call sites and fix the SELECT path.

python/cudaq_algorithms/_qsvt.py
```diff
@@ -0,0 +1,219 @@
...
+                                    walk_directions=None,
+                                    convention=None):
+    """Pack QSVT phase data and PauliLCU layout for apply_phase_sequence()."""
+
+    sequence = _phase_sequence(phases, walk_directions, convention)
+    angles, term_controls, term_ops, term_lengths, term_signs = (
+        _pauli_lcu_kernel_tuple(kernel_data))
+    return (sequence.phase_data, sequence.walk_direction_data, list(angles),
+            list(term_controls), list(term_ops), list(term_lengths),
+            list(term_signs))
```
kvmto: **[Blocking] `PhaseSequence.convention` is silently ignored by the device-execution path.**

A sequence tagged `qsvt.phase_sequence(phases, convention="qsp")` behaves inconsistently between the two official consumers:

- `qsvt.phases_to_poly(seq)` reads `seq.convention` and applies the correct per-convention 2×2 math.
- `seq.kernel_data()` / this function → `qsvt.apply_phase_sequence` never inspect `convention` and forward the raw phases as projector-convention values.

These are not equivalent up to global phase: `diag(e^{iφ}, 1)` is not a scalar multiple of `diag(e^{iφ}, e^{−iφ})` for φ ≠ 0 (verified algebraically and numerically). So the tagging pattern — which is tested (`test_qsvt.py:676-680`) and looks like the supported way to carry QSP phases — produces a genuinely wrong unitary on device with no warning. Every in-repo caller happens to dodge this by manually calling `projector_phases_from_qsp()` *before* building the sequence. Note also that the correct QSP-native C++ kernel (`qsvt_primitives::apply_qsp_sequence`) exists but is never Python-bound, so there is currently no correct execution path for a qsp-tagged sequence at all.

Suggested fix (in preference order): (a) auto-apply the ×2 conversion here / in `kernel_data()` when `convention == qsp`; (b) raise on non-`qsvt` conventions at the device boundary; (c) at minimum, document loudly that `convention` is diagnostic-only.

include/cudaq/algorithms/qsvt/qsvt.h
```diff
@@ -0,0 +1,739 @@
...
+
+  apply_qsvt_signal_phase(signal, phases[0]);
+  for (std::size_t i = 1; i < phases.size(); ++i) {
+    if (direction == qsvt_walk_direction::forward)
+      apply_qubitization_walk(signal, system, encoding);
+    else
+      apply_adjoint_qubitization_walk(signal, system, encoding);
+    apply_qsvt_signal_phase(signal, phases[i]);
+  }
+}
```
kvmto, line 201: **[Blocking] This composition implements the wrong unitary for any ≥2-term Hamiltonian.**

`apply_qsvt_signal_phase` (line 73) applies the projector phase on the **computational |0…0⟩** ancilla state, but `apply_qubitization_walk` reflects about the **PREPARE state |φ⟩** (`reflect_about_prepare`). QSVT requires the phase projector and the walk reflection to be defined w.r.t. the *same* signal subspace; they coincide only when |φ⟩ = |0⟩, i.e. the degenerate single-term / 0-ancilla case — which happens to be the only case the C++ tests check numerically.

Two independent from-scratch gate-level simulations (12/12 cases: three 2-term Hamiltonians, degree-1 and degree-3 phase sequences, both eigenvalues) show the good-subspace amplitude matches `evaluate_qsvt_response(phases, ±x)` in **neither sign, with the magnitude also wrong** (e.g. 0.166 vs the correct 0.721) — so this is a different unitary, not a sign/phase-convention ambiguity. The composition becomes exactly correct (reproducing `evaluate_qsvt_response(phases, −x)`, bit-for-bit identical to the Python-facing `qsvt::apply_phase_sequence` path) with two changes: (1) the signal phase applied as a projector on |φ⟩ (i.e. `unprepare; |0⟩-phase; prepare`), and (2) the ancilla prepared into |φ⟩ before the first phase.

This entire family (`apply_qsvt_sequence`, `apply_qsp_sequence`, controlled variants) is currently called from nowhere except `test_qsvt.cpp`, so shipped results are unaffected today — but this header is the documented C++ entry point for QSVT-over-`pauli_lcu`. Please either fix the composition or delete/deprecate this family in favor of the single verified implementation in `kernels.cpp`.

lib/block_encoding/device/kernels.cpp
```diff
@@ -0,0 +1,409 @@
...
+}
+
+__qpu__ void apply_sequence(cudaq::qview<> signal, cudaq::qview<> system,
+                            const std::vector<double> &phases,
+                            const std::vector<int> &walk_directions,
+                            const std::vector<double> &state_prep_angles,
+                            const std::vector<int> &term_controls,
+                            const std::vector<int> &term_ops,
+                            const std::vector<int> &term_lengths,
+                            const std::vector<int> &term_signs) {
```
kvmto: **[Blocking — same defect as the `qsvt.h` sequence family]** `qsvt_primitives::apply_sequence`/`apply_qsp_sequence` compose `qubitization::apply_walk` (SELECT + `reflect_about_prepare`) with `apply_signal_phase` (computational-|0⟩ projector) — structurally identical to the wrong composition in `qsvt.h`, and distinct from the *correct* `cudaq_algorithms::qsvt::apply_phase_sequence` just below (which pairs the full block encoding with `reflect_about_zero`, and was verified to match `evaluate_qsvt_response(phases, −x)` exactly). Nothing calls these today, but they're exported through `kernels.h` and are one `addDeviceKernelInterop` call away from Python exposure. Suggest removing them or fixing them together with `qsvt.h`.

lib/unittests/test_qsvt.cpp
```diff
@@ -0,0 +1,290 @@
...
+  };
+  EXPECT_NO_THROW(controlled_qsp_sequence_functor_test());
+}
+
+// Test purpose: verify QSVT sequences reproduce one and two qubitization walks.
+TEST(QSVTTester, qsvt_sequence_executes_expected_walk_powers) {
+  using namespace cudaq::spin;
+  using namespace cudaq::algorithms;
+
+  cudaq::spin_op h = x(0);
```
kvmto, line 220: **[Blocking — this is the regression-guard gap that let the `qsvt.h` composition bug through.]** This is the only test in the file that asserts numerics via `get_state`, and it uses `h = x(0)` — a single Pauli term → `num_ancilla == 0`, where `prepare`/`unprepare` are literal no-ops (`if (ancilla.size()==0) return;`) and the reflections vanish through the arity macros. In that regime the walk collapses to a bare X gate, forward and adjoint walks are gate-for-gate identical, and phase placement is unobservable — so this test cannot detect the wrong-projector composition flagged in `qsvt.h`. Meanwhile every test above that uses the real 2-term `0.5*x(0)+0.3*z(0)` asserts only `EXPECT_NO_THROW`.

Please add ≥2-term numeric tests for `apply_qsvt_sequence`/`apply_qsp_sequence` (and controlled variants), asserting `get_state` output against `evaluate_qsvt_response(phases, −eigenvalue)` — the same methodology `tests/python/test_qsvt.py:404-444` already uses for the Python path. That single test class would have caught both the composition bug and the sign-drop bug.

lib/unittests/test_qubitization.cpp
```diff
@@ -0,0 +1,327 @@
...
+  EXPECT_NO_THROW(controlled_adjoint_walk_power_functor_test());
+}
+
+// Test purpose: verify controlled walk execution respects the control state.
+TEST(QubitizationTester, checkControlledWalkExecution) {
+  using namespace cudaq::spin;
+  using namespace cudaq::algorithms;
+
+  cudaq::spin_op h = x(0);
+  pauli_lcu encoding(h, 1);
```
kvmto, line 278: **[Test coverage — same degenerate pattern as `test_qsvt.cpp`.]** Both numeric-assertion tests in this file (`checkControlledWalkExecution`, `checkControlledWalkPowerExecution`) use `h = x(0)` (0 ancilla), where PREPARE/UNPREPARE conjugation, the multi-controlled reflections, and walk ordering are all no-ops or trivial; every ≥2-term test asserts only `EXPECT_NO_THROW`. A review-side simulation of a non-degenerate 4-term/2-ancilla encoding confirmed the walk, adjoint-power semantics, and observable builders are currently *correct* — but none of that is protected by the suite. Suggest: (a) verify the reflection-observable expectation after PREPARE + k walks reproduces T₂ₖ(H/α) for a 2-term H, and (b) verify `apply_adjoint_qubitization_walk_power(n)` undoes `apply_qubitization_walk_power(n)` for n ≥ 2, non-degenerately, including controlled variants.

tests/python/test_qsvt.py
```diff
@@ -0,0 +1,752 @@
...
+    return _zero_ancilla_component(full_state, num_system, num_ancilla)
+
+
+def _qsp_to_projector_phases(phases):
+    return algorithms.qsvt.projector_phases_from_qsp(phases)
+
+
+def _qsppack_hamiltonian_simulation_phases(tau, degree=16):
+    qsppack = pytest.importorskip("qsppack")
+    scipy_special = pytest.importorskip("scipy.special")
```
kvmto, line 179: **[CI gap] These `importorskip` calls mean the strongest test in the suite never runs in CI.** `test_qsvt_hamiltonian_simulation_using_qsppack` — the only test comparing device QSVT Hamiltonian simulation against exact `exp(−iHt)` with dynamically generated phases — depends on `qsppack`, which is a real PyPI package but is never installed by `.github/workflows/lib_algorithms.yaml` (it installs only `cmake ninja nanobind numpy pytest`) or any requirements file. The test silently reports *skipped* and CI stays green. Please add `qsppack` (and `scipy`) to the CI install step, and consider a CI step that executes the `hamiltonian_simulation` examples end-to-end.

examples/quantum_exact_lanczos/quantum_exact_lanczos_molecules.py
```diff
@@ -0,0 +1,431 @@
...
+
+
+def comparison_energy(data: QubitHamiltonianData,
+                      exact_max_qubits: int) -> tuple[float | None, str]:
+    """Choose dense exact diagonalization or stored reference data."""
+    if data.num_qubits <= exact_max_qubits:
+        return exact_ground_energy(data), "dense exact diagonalization"
+    if data.reference_energy is not None:
+        return data.reference_energy, data.reference_energy_kind
+    return None, "none"
```
kvmto, line 195: **[Correctness of the demo comparison]** Both `n2_active_space_jw.json` and `benzene_active_space_jw.json` have `num_qubits == 8`, and `DEFAULT_EXACT_MAX_QUBITS == 8`, so with `<=` this always takes the dense-diagonalization branch and the stored higher-precision CASCI `reference_energy` those files ship with is unreachable under default settings. Because these two fixtures' Pauli coefficients are rounded to ~6 significant digits, the re-diagonalized value differs from the stored CASCI reference by ~2–3e-4 Ha (verified for both) — masked only by the loose 5e-2 default tolerance. Suggest `<` instead of `<=` (or documenting the intent), and adding at least one test that runs the actual quantum moment workflow for an 8-qubit fixture (only H2's 4-qubit path is currently exercised).

examples/quantum_exact_lanczos/quantum_exact_lanczos_h2.py
```diff
@@ -0,0 +1,159 @@
...
+        print(f"  ... {len(data.terms) - max_terms} more terms")
+
+
+def print_matrix(name: str, matrix: np.ndarray):
+    """Print a named dense matrix with compact numerical formatting."""
+    print(f"{name} shape: {matrix.shape}")
+    print(np.array2string(matrix, precision=8, suppress_small=True))
+
+
+def main() -> int:
```
kvmto, line 61: **[Test coverage]** Both READMEs call this script "the recommended starting point," and it re-implements (rather than calls) the tested `run_qel_workflow()` arithmetic — its own energy reconstruction, tolerance check, and argument handling are never executed by any test (`test_qel_example.py` only loads `quantum_exact_lanczos_molecules.py`). A bug introduced in this copy would ship undetected to the first file new users run. Suggest an importlib-based pytest mirroring the one that covers `quantum_exact_lanczos_molecules.py`.

lib/unittests/test_krylov.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+  EXPECT_EQ(required_chebyshev_moments(0), 0);
+  EXPECT_EQ(required_chebyshev_moments(1), 2);
+  EXPECT_EQ(required_chebyshev_moments(4), 8);
+}
+
+TEST(KrylovTester, checkChebyshevMatrixConstruction) {
+  using namespace cudaq::algorithms::krylov;
+
+  const std::vector<double> moments = {1.0, 0.2, -0.1, 0.05};
```
kvmto, line 24: **[Nit — defense-in-depth]** The expected S/H values here are exactly what the code's own product-to-sum formula produces for these hand-picked moments, so the test is self-consistent rather than physically grounded. The formulas themselves were independently re-derived and verified (moments from an explicit Hermitian matrix reproduce its exact eigenvalues via `eigh(H, S)` to ~1e-16), and real physical validation exists in `test_qel_example.py` — but a colocated test that builds moments from an explicit small operator and checks the generalized eigenproblem against `eigvalsh` would guard this primitive directly.

kvmto (changes requested):
## Review summary — Request changes

Thanks Scott — this is a big, well-structured PR, and the core physics design is sound. The LCU PREPARE/SELECT machinery, the host-side QSP/QSVT response math, and the Chebyshev-moment/Krylov method were each verified against independent gate-level simulation (and, for the two most serious findings below, by compiling this branch and checking `get_state` output against dense references). Those all check out.

However, the review found **two confirmed silent wrong-answer bugs reachable from the shipped Python API**, **one confirmed wrong-answer bug in the primary documented C++ header**, and a systematic test-coverage pattern that made all three invisible. Details are in the inline comments; short index:

**Blocking**
1. `PauliLCU::select`/`apply` silently encodes `+H` instead of `-H` for single-term negative-coefficient Hamiltonians (missing `COUNT==0` branch in the Z-arity macro). Verified by compiled execution: `apply(+0.5·X0Z1)` and `apply(-0.5·X0Z1)` produce bit-for-bit identical states. → comments on `lib/block_encoding/device/kernels.cpp` and `include/cudaq/algorithms/detail/qpu_dispatch.h`.
2. `PhaseSequence.convention="qsp"` is honored by `phases_to_poly` but silently ignored by the device-execution path — a tagged-then-executed sequence runs a genuinely different unitary (not a global phase). → comment on `python/cudaq_algorithms/_qsvt.py`.
3. `qsvt.h`'s `apply_qsvt_sequence`/`apply_qsp_sequence` family composes a computational-|0⟩ signal phase with a PREPARE-state-reflection walk; for any ≥2-term Hamiltonian this is a different unitary than `evaluate_qsvt_response` predicts (12/12 simulated cases wrong, magnitude included). Currently dead code, but it is the documented C++ entry point, and an identical exported duplicate exists in `qsvt_primitives::`. → comments on `include/cudaq/algorithms/qsvt/qsvt.h` and `lib/block_encoding/device/kernels.cpp`.
4. Test coverage: every numeric (`get_state`) C++ test of the walk/QSVT path uses a degenerate single-term (0-ancilla) Hamiltonian where PREPARE/reflections are no-ops; every ≥2-term test asserts only `EXPECT_NO_THROW`. This is exactly why 1 and 3 shipped. → comments on `lib/unittests/test_qsvt.cpp` and `test_qubitization.cpp`.

**Non-blocking (please still address)**
5. The only exact-`exp(-iHt)` validation test is permanently *skipped* in CI (`qsppack` is never installed anywhere).
6. `quantum_exact_lanczos_h2.py` — the READMEs' "recommended starting point" — has zero test coverage.
7. The N2/benzene fixtures' stored CASCI references are unreachable under default settings (`num_qubits == 8` and `<=` vs `DEFAULT_EXACT_MAX_QUBITS = 8`).
8. Krylov unit tests assert the formula's own output for hand-picked moments (tautological); the math itself was verified correct independently.

**Verified correct (for reviewer confidence, no action needed):** LCU bit-ordering/padding and sign-branch placement (3-, 5-, and 10-term cases, simulation + compiled execution); `checkBlockEncodingMatchesDenseHamiltonianAction`; host-side QSVT math incl. the QSPPACK cosine reference (~1e-9); the intentional `reflect_about_zero` vs `build_qubitization_reflection_observable` sign asymmetry (cancels in all Chebyshev-moment expectations, checked against exact T_n(H/α) for n=0..5); `build_lcu_select_observable` register offsets; adjoint-walk-power semantics ((Wⁿ)† = (W†)ⁿ looped, verified non-degenerately); Krylov S/H matrix formulas (reproduce exact spectra to 1e-16).

**Suggested merge path:** fix 1–2 → fix or deprecate 3 (and the `qsvt_primitives` duplicates) → add the non-degenerate numeric tests from 4 as the regression guard → wire `qsppack` into CI → 6–8 as fast-follow.

lib/block_encoding/device/kernels.cpp
wsttiger: Fixed in 9305097. `select` now handles the 0-ancilla negative-coefficient case explicitly: `rz(2*pi)` on `system[0]`, which is exactly the −I matrix (so it also stays correct under control synthesis), instead of relying on the arity ladder. Added the dense-reference regression test `checkSingleTermNegativeCoefficientEncodesSign` (single-term `-0.5*X0Z1` vs dense Pauli action), which fails without the fix.

include/cudaq/algorithms/detail/qpu_dispatch.h
wsttiger: Fixed in 9305097. The COUNT==0 comment is rescoped to the reflection and signal-phase call sites where the dropped phase really is global, and now explicitly notes that `block_encoding::select` handles its 0-ancilla sign case itself. While auditing call sites I also found the header `controlled_reflect_about_zero` had an `if (num_ancilla > 0)` guard that skipped the COUNT==0 `z(control)` branch the device-kernel copy applies (a real relative phase on the control) — removed in the same commit.

python/cudaq_algorithms/_qsvt.py
wsttiger: Fixed in e02f83b, taking option (a): `_PhaseSequence.projector_phase_data()` doubles qsp-tagged phases, and both `kernel_data()` and `pauli_lcu_kernel_args` now pack through it, so a qsp-tagged sequence executes the equivalent projector circuit. `phase_data` stays raw so `phases_to_poly` keeps honoring the convention itself. Added `test_qsvt_qsp_convention_kernel_packing`, which checks both the packed arguments and end-to-end statevector equality against manually converted phases.

include/cudaq/algorithms/qsvt/qsvt.h
wsttiger: Fixed in 4466d91. The whole sequence family is rewritten to the verified composition: each step is the full block encoding (PREPARE, SELECT, PREPARE†) followed by `reflect_about_zero` on the signal register, so the phase projector and the walk reflection act on the same |0…0⟩ subspace — matching `qsvt::apply_phase_sequence`. The signal register now starts at |0…0⟩ and callers no longer prepare it (docs and all call sites updated). Controlled steps wrap a controlled SELECT in uncontrolled PREPARE/PREPARE† plus the controlled zero reflection, so a controlled sequence collapses to the identity at control |0⟩. 558ae46 adds the numeric coverage: good-subspace amplitudes match `evaluate_qsvt_response(phases, -eig/alpha)` on a 2-term H for degree-1 and degree-3 sequences, and the header family matches the device `qsvt::apply_phase_sequence` bit for bit (including an adjoint step).

lib/block_encoding/device/kernels.cpp
wsttiger: Deleted in 4466d91 rather than fixed — `qsvt_primitives::apply_sequence`/`apply_qsp_sequence` had no callers outside their own compile tests and duplicated what the correct `qsvt::apply_phase_sequence` already provides. The signal-phase primitives remain.

lib/unittests/test_qsvt.cpp
wsttiger: Addressed in 558ae46. New numeric tests on `0.5*x(0) + 0.3*z(0)` (1 ancilla, eigenvalues ±sqrt(0.34), system prepared in the ry(atan2(0.5,0.3)) eigenstate): good-subspace amplitudes must equal `evaluate_qsvt_response(phases, -eigenvalue/alpha).value * eigenvector` for odd degree-1 and degree-3 phase sets (odd degree keeps the −H/alpha sign observable, with an explicit +eigenvalue must-NOT-match guard), plus qsp(phi) ≡ qsvt(2*phi) statevector equality, controlled on/off behavior, and bit-for-bit equality against the device `qsvt::apply_phase_sequence` path. These fail against the old composition.

lib/unittests/test_qubitization.cpp
wsttiger: Addressed in 558ae46. Two non-degenerate execution tests: (a) PREPARE + walk^k + UNPREPARE reproduces ⟨T_2k(H/alpha)⟩ via the ancilla reflection observable for k=1..3 on `0.2*I + 0.5*X + 0.3*Z` — chosen for its asymmetric spectrum 0.2±sqrt(0.34) so the moment genuinely depends on both eigenvalue weights; (b) `apply_adjoint_qubitization_walk_power(n)` undoes `apply_qubitization_walk_power(n)` for n=2,3 on a 3-term, 2-ancilla, 2-system-qubit H, including the controlled variants, with full statevector comparison against a preparation-only reference.

tests/python/test_qsvt.py
wsttiger: Fixed in b7c9198 — the CI pip install now includes `scipy qsppack`, so `test_qsvt_hamiltonian_simulation_using_qsppack` actually runs (verified locally with both installed: it passes). Running the hamiltonian_simulation examples end-to-end in CI is a good follow-up; left out of this PR to keep it focused on the review fixes.

examples/quantum_exact_lanczos/quantum_exact_lanczos_molecules.py
wsttiger: Fixed in e8e6878 — boundary changed to `<` so 8-qubit fixtures use their stored CASCI references, with the semantics documented in the `comparison_energy` docstring and the `--exact-max-qubits` help text. The same commit adds `test_n2_measured_low_order_moments_match_exact_chebyshev`, which runs the actual 8-qubit quantum moment workflow (the occupied-qubits (0,1,2,3) path) and compares dimension-2 moments against exact dense Chebyshev values.

examples/quantum_exact_lanczos/quantum_exact_lanczos_h2.py
wsttiger: Addressed in e8e6878. `test_h2_walkthrough_example_matches_dense_exact` loads the script via importlib (reusing the molecules-module fixture so its sibling import resolves) and runs `main()` end to end on qpp-cpu — which exercises its argument handling, energy reconstruction, and internal tolerance check (`main()` raises if the QEL energy misses dense exact diagonalization).

lib/unittests/test_krylov.cpp
wsttiger: Addressed in e8e6878 with the suggested defense-in-depth on the Python side: `test_chebyshev_matrices_reproduce_exact_eigenvalues` builds moments mu_k = v^T T_k(A) v from an explicit 3x3 Hermitian matrix (spectrum scaled inside [-1,1]), feeds them through `krylov.build_chebyshev_matrices`, solves the generalized eigenproblem by symmetric whitening, and requires the eigenvalues to match `numpy.linalg.eigvalsh(A)` to 1e-8.

wsttiger:
Thanks @kvmto for the thorough review — all 11 comments are addressed in six commits, replied to inline:

- 2f6670c — Fix dropped sign for single-term LCU SELECT (explicit `rz(2π)` = −I for the 0-ancilla case, qpu_dispatch comment rescoped, header `controlled_reflect_about_zero` guard removed, dense regression test)
- 37efdd5 — Convert qsp-tagged phases for device QSVT execution (option (a): auto-double in kernel packing, raw phases kept for `phases_to_poly`)
- 47080ba — Fix QSVT composition in the qsvt.h sequence family (block encoding + zero reflection per step, signal starts at |0…0⟩, callers no longer prepare; broken `qsvt_primitives` sequence kernels deleted)
- ecb7d09 — Add non-degenerate QSVT and qubitization execution tests (host-response match on eigenstates for degrees 1/3 with sign guard, qsp ≡ 2φ, controlled on/off, bit-for-bit device-path equality; Chebyshev moments on an asymmetric-spectrum 3-term H, adjoint walk powers invert walk powers incl. controlled)
- 58fdf8e — Install scipy and qsppack in CI (QSPPACK test now runs)
- 6d16225 — Fix QEL exact-diagonalization boundary (`<=` → `<`) and add example tests (8-qubit N2 moment path, h2 walkthrough end-to-end, krylov explicit-matrix eigenvalue test)

Verified locally: 71/71 C++ ctest and the full `tests/python` suite (49 passed; QSPPACK test included with `scipy`/`qsppack` installed — the only skip is the pre-existing pyscf one).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

*(Edit: commit hashes updated after a history rewrite on the branch; content is unchanged — trees are byte-identical.)*

PR #11, opened 2026-07-02
Add PR checks and cache handling

anjbur:
Expanded work in #10 to set up CI checks for PRs. Specifically:

- Add PR checks to build any of: CUDA-Q, the algorithms library, or the algorithms wheel, depending on the files changed
- Installed `copy-pr-bot` and set up PR workflow with related triggers, so that PRs from forks can use NVIIA runners
- Replace `ci.yaml` with PR sanity checks
- Expand the lib_algorithms matrix to amd64/arm64 across CUDA 12.6/13.0
- Add CUDA-Q cache initialization and cleanup
- Sync `get-cudaq-wheels` action with updates in cudaqx, so that it doesn't rebuild on a cache hit
- Containerized the algorithms metapackage build

Testing was done in the CI to validate. For most recent run that included the CUDA-Q build, see [the build](https://github.com/NVIDIA/cudaq-algorithms/actions/runs/28625759759/job/84891769286) against [4cd87d8](https://github.com/NVIDIA/cudaq-algorithms/pull/11/commits/4cd87d8472bb316f3fbfd7152ae22d12108b71bf). 

The workflows `cudaq_cache.yaml` and `pr_cache_cleanup.yaml` were not run, as they don't exist yet on main and can't be triggered with `workflow_dispatch`. However, they are identical to the same-named cudaqx workflows.


.github/workflows/pr_workflow.yaml
```diff
@@ -0,0 +1,243 @@
...
+        uses: ./.github/actions/get-cudaq-version
+
+      - name: Get CUDAQ build
+        uses: ./.github/actions/get-cudaq-build
+        with:
+          repo: ${{ steps.get-cudaq-version.outputs.repo }}
+          ref: ${{ steps.get-cudaq-version.outputs.ref }}
+          token: ${{ secrets.CUDAQ_ACCESS_TOKEN }}
+          cache-key-suffix: pr${{ needs.check-changes.outputs.pr-number }}
+          save-build: false
```
wsttiger: On a full cache miss this falls through to `get-cudaq-build`'s **Build CUDAQ** step (its only gate is `check-cache.valid == 'false'`), i.e. a cold CUDA-Q source build in all 4 matrix cells on cpu8 runners — and with `save-build: false` the result is never cached, so it repeats on every push to every open PR.

The trigger is realistic: GitHub deletes caches not accessed for 7 days (the `ccache-cudaq-*` entries this repo's cache workflow saves are already absent from `gh cache list`). After a quiet week, the first push to any PR that doesn't touch the cudaq filter paths hits this, and nothing self-heals — `cudaq_cache.yaml` has no `schedule` trigger, so recovery requires a manual dispatch.

Suggestions: set `save-build: true` here (the pr-suffixed key gets cleaned up by `pr_cache_cleanup` anyway), and/or add a weekly `schedule` to `cudaq_cache.yaml` to keep the main caches warm past the 7-day expiry.

.github/workflows/cudaq_cache.yaml
```diff
@@ -0,0 +1,91 @@
...
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+    paths:
+      - '.github/workflows/cudaq_cache.yaml'
+      - '.github/actions/get-cudaq-build/**'
+      - '.github/actions/get-cudaq-wheels/**'
+      - '.cudaq_version'
```
wsttiger, line 15: This paths list omits `.github/actions/get-cudaq-version/**`, which `pr_workflow.yaml`'s `build-cudaq` filter *does* include. The cache key embeds the resolved ref (`cudaq-<platform>-<ref>-<hash>`, and the hash only covers `get-cudaq-build/**` + `.cudaq_version`), so a merged change to `get-cudaq-version` that alters the resolved ref (defaults, fallback logic, jq keys) produces a new key that no main-branch run ever populates — the PR's own `-prN` cache is deleted on branch delete, and every subsequent PR lands in the uncached rebuild path until someone manually dispatches this workflow.

Adding `- '.github/actions/get-cudaq-version/**'` here keeps the two filters symmetric.

.github/workflows/pr_sanity_checks.yaml
```diff
@@ -0,0 +1,202 @@
...
+# All rights reserved.                                                         #
+#                                                                              #
+# This source code and the accompanying materials are made available under     #
+# the terms of the Apache License 2.0 which accompanies this distribution.     #
+# ============================================================================ #
+
+name: PR sanity checks
+
+on:
+  pull_request:
```
wsttiger, line 12: The deleted `ci.yaml` also listened on `merge_group` (checks_requested) and `workflow_dispatch`; neither survives here or in `pr_workflow.yaml`. That's latent today (no merge queue / required-check ruleset is configured on main), but if a merge queue is ever enabled with these checks required, `merge_group` events trigger nothing, the required checks never report, and every queue entry stalls. Losing `workflow_dispatch` also means no manual re-runs. Worth re-adding both triggers while this file is fresh.

.github/workflows/cudaq_cache.yaml
```diff
@@ -0,0 +1,91 @@
...
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
```
wsttiger: For a cache-warming workflow, `cancel-in-progress: true` seems like the wrong default: two merges touching `.cudaq_version` (or the actions) landing close together cancel the first run's 4-cell CUDA-Q build matrix on the cpu32 runners mid-build — hours of runner time burned with no cache saved — and the second run restarts from scratch. Queueing (`cancel-in-progress: false`) wastes nothing here since runs are rare and each one's output is what every PR depends on.

wsttiger (commented):
Nice CI architecture overall — I went through the new workflows, the composite-action interfaces, the cache-key/cleanup suffix contract, and the concurrency-group rename, and those all check out. A few things I also chased and confirmed are fine (so nobody else has to): the forced `CC=gcc`/`CXX=g++` plus devdeps gcc-toolset-12 sourcing in `build_wheels.sh` is the right call (the devdeps image pre-sets a stale CC/CXX), the `mkdir -p`/`chown` change matches the cudaqx wheel-cache-restore fix, and the `ubuntu:24.04` container on the metapackage job fixes a real root-`mkdir /metapackages` permission failure.

Inline comments below — the two I'd weigh most are the cache-lifecycle ones (uncached CUDA-Q rebuilds on cpu8 after cache expiry, and the `get-cudaq-version` trigger-path gap), since both fail quietly and expensively. One small suggestion for `wheel_algorithms.yaml`: a one-line comment on the `-build-wheels-` concurrency infix explaining the reusable-workflow collision would keep someone from "normalizing" it back later.

.github/workflows/pr_workflow.yaml
```diff
@@ -0,0 +1,243 @@
...
+        with:
+          repo: ${{ steps.get-cudaq-version.outputs.repo }}
+          ref: ${{ steps.get-cudaq-version.outputs.ref }}
+          token: ${{ secrets.CUDAQ_ACCESS_TOKEN }}
+          cache-key-suffix: pr${{ needs.check-changes.outputs.pr-number }}
+          save-build: true
+          save-ccache: false
+          platform: ${{ matrix.platform }}-cu${{ steps.config.outputs.cuda_major }}
+
+  build-algorithms:
```
wsttiger, line 100: This job re-implements `lib_algorithms.yaml`'s `build-and-test` inline (~60 lines: same matrix, container, `get-cudaq-build` call, installs, `test_algorithms_build.sh` invocation), even though `lib_algorithms.yaml` is already `workflow_call`-able with a `cache_key_suffix` input — and this same file already uses that pattern for wheels (`uses: ./.github/workflows/wheel_algorithms.yaml`). The two copies already disagree (`save-build: true` there vs `false` here; `$GITHUB_OUTPUT` quoting), and this PR had to edit the matrix in both.

One caveat before switching to `workflow_call`: `lib_algorithms.yaml` still has `group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}`, so calling it from here would hit the exact self-cancellation collision you just fixed for `wheel_algorithms.yaml` — it needs the same group rename first.

.github/workflows/cudaq_cache.yaml
```diff
@@ -0,0 +1,91 @@
...
+
+jobs:
+  build-cudaq:
```
wsttiger, line 22: `build-cudaq` here is a near-verbatim copy of `pr_workflow.yaml`'s `build-cudaq`, and `build-cudaq-wheels` below likewise copies `wheel_algorithms.yaml`'s (down to the cu13 matrix comment). Since these copies feed the same cache keys from different files, drift is worse than cosmetic (they already differ: backticks vs `$(...)` in the Configure step): bump the devcontainer tag or matrix in one file but not the other and the main cache gets populated for a different platform/toolchain set than PR jobs restore, silently defeating the cache. A small reusable workflow parameterized by `save-build`/`save-ccache`/`cache-key-suffix` would give both callers one definition.

.github/workflows/pr_workflow.yaml
```diff
@@ -0,0 +1,243 @@
...
+              - 'cmake/**'
+              - 'include/**'
+              - 'lib/**'
+              - 'python/**'
+              - 'tests/**'
+              - 'scripts/test_algorithms_build.sh'
+              - 'scripts/validation/**'
+              - 'pyproject.toml.cu*'
+            build-examples:
+              - 'examples/**'
```
wsttiger: There's no `examples/` directory in the repo yet, and no job builds examples — this output only feeds one OR-clause gating `build-algorithms`, which never compiles anything under `examples/`. So a future PR adding a broken example goes green while the filter suggests examples are covered. Until an examples job exists, folding `examples/**` into the `build-algorithms` filter list (or dropping it) avoids the false signal.

.github/workflows/pr_sanity_checks.yaml
anjbur: Personally, I think it's better to leave off the merge_group triggers unless we have imminent plans to enable merge groups in this repo. Otherwise, it's unused wiring that makes it confusing whether merge groups are present in the repo not.

On the workflow_dispatch trigger: I removed it from this and the pr_workflow.yaml workflows because neither function outside of the context of a pull request, due to how the path filtering works. The existing triggers will already run the job any time a PR is opened or updated, and those jobs can be manually re-run without the workflow_dispatch trigger.

.github/workflows/pr_workflow.yaml
anjbur: Good catch, I'll remove this filter path

.github/workflows/pr_workflow.yaml
anjbur: I'll switch to a lib_algorithms call and normalize the cache usage

.github/workflows/pr_workflow.yaml
anjbur: That's a fair point, about the cache expiry. I was thinking that this shouldn't ever result in a cache miss, as `build-cudaq` was set as a dependency and would save the cache for us on changes. Easy to change this to save the build though!

.github/workflows/cudaq_cache.yaml
anjbur: Agreed, I'll add that path

.github/workflows/cudaq_cache.yaml
anjbur: This workflow on the cudaqx repo typically takes ~5-6 minutes for a CUDAQ build and ~20-25 minutes for a CUDAQ wheel build, so at the very least it shouldn't ever waste hours of runner time! My only concern here is from the workflow_dispatch trigger, rather than the automatic triggers (which I agree, should be infrequent enough to not be an issue). Having `cancel-in-progress: true` limits the load that manual runs can pass to the NVIDIA runners. 

Another option is to change the concurrency group to `${{ github.workflow }}-${{ github.sha }}`. That way the same branch could have multiple simultaneous runs, but not the same commit

.github/workflows/cudaq_cache.yaml
anjbur: Agreed about the duplication. I'll take a look and see what we can extract out

.github/workflows/cudaq_cache.yaml
anjbur: Done

.github/workflows/pr_workflow.yaml
anjbur: Done

.github/workflows/pr_workflow.yaml
anjbur: Done by switching to a `lib_algorithms.yaml` call, which caches the CUDA-Q build

.github/workflows/cudaq_cache.yaml
anjbur: Done

.github/workflows/pr_workflow.yaml
anjbur: Done

.github/workflows/cudaq_cache.yaml
anjbur: I've thought about this a bit more and decided I agree with you here, rather than me. Manually runs shouldn't be frequent enough to create a problem. Switched to `cancel-in-progress: false`

wsttiger (approved):
LGTM

PR #12, opened 2026-07-08
Add LCU, qubitization, and QSVT primitives

wsttiger:
## Summary

This PR adds the first three quantum primitives to `cudaq_algorithms`:
**Pauli LCU block encoding**, **qubitization walks** with **Chebyshev moment**
measurement, and **Quantum Singular Value Transform(QSVT)** phase sequences — each with controlled variants.
It also introduces the **`BlockEncoding` protocol**, so the walk and QSVT
operators are generic over the encoding from day one rather than hard-wired
to the LCU implementation.

12 commits, 3,052 insertions across 17 files. No changes to the existing
library or bindings.

## The primitives

```python
from cudaq_algorithms import PauliLCU, Walk, QSVT, PhaseSequence

enc = PauliLCU({"ZI": 0.70, "IZ": -0.43, "XX": 0.19})   # dict, SpinOperator,
enc.alpha, enc.num_ancilla, enc.terms                    # or (coeff, word) pairs

walk = Walk(enc)
walk.kernel(power=3)                  # PREPARE + W^3 + UNPREPARE
walk.controlled_kernel(power=2)       # controlled walks, both control states
walk.moment(psi, k)                   # <T_k(H/alpha)> via cudaq.observe

qsvt = QSVT(enc)
seq = PhaseSequence(phases, convention="qsp")    # qsvt/qsp conventions
qsvt.kernel(seq)                                 # @cudaq.kernel(state)
```

Every object is a kernel *factory*: construct the primitive once,
parameterize at kernel-request time. Statevector-only conveniences
(`action`, `transform`, `good_subspace`) are isolated in
`cudaq_algorithms.sim_utils` — nothing in the primitive modules calls
`cudaq.get_state`, and `Walk.moment` measures through `cudaq.observe`, which
is a hardware-legitimate path.

## The `BlockEncoding` protocol (dependency injection)

`Walk` and `QSVT` consume encodings only through a structural,
runtime-checkable protocol (`block_encoding.py`): register geometry,
`alpha`, an odd-moment observable hook, and eight kernel factories whose
returned kernels have **fixed, data-free signatures** — all
encoding-specific data is captured inside the kernel at factory time. This
relies on a verified CUDA-Q behavior: a `@cudaq.kernel` can call kernels
captured through closure variables, including factory-made kernels invoked
with register views inside loops.

`PauliLCU` conforms structurally (deliberately no inheritance — pinned by a
test). Future encodings (e.g. double-factorized) plug in by satisfying the
same surface, with no changes to `Walk`/`QSVT` or to user call sites.

Module layout matches the architecture:

```
block_encoding.py  the protocol Walk and QSVT consume
common_kernels.py  encoding-independent kernels (zero reflections, signal phases)
pauli_lcu.py       everything LCU: PauliLCU + prepare/select/apply,
                   walk-step and phase-sequence kernels, SELECT observable
qubitization.py    Walk (moments) + reflection observable — encoding-generic
qsvt.py            PhaseSequence, QSVT — encoding-generic
sim_utils.py       simulation-only helpers (statevector access)
```

The module-level kernels (`prepare`, `select`, `apply`,
`apply_phase_sequence`, ...) remain exported as the escape hatch for
composing inside user kernels, fed by `PauliLCU.kernel_args`.

## Testing

**37 tests**, all passing (CUDA-Q 0.15, `qpp-cpu`; suite honors
`CUDAQ_DEFAULT_SIMULATOR` — an fp64 target is required for the tolerances):

- `test_pauli_lcu` / `test_qubitization` / `test_qsvt` (23): dense-reference
  correctness — Pauli-sum action, Chebyshev moments on an asymmetric
  spectrum, adjoint walks inverting forward walks, controlled circuits
  against uncontrolled references at both control states, the QSVT
  good-subspace block against the 2x2 signal model column by column, qsp/qsvt
  convention equivalence, and a QSPPACK-driven Hamiltonian simulation
  reaching ~1e-15 state error.
- `test_block_encoding_protocol` (5): a foreign encoding class (no PauliLCU
  inheritance, no `kernel_args`) drives Walk/QSVT to results identical to
  the PauliLCU reference — the consumers cannot reach around the protocol.
- `test_walk_qsvt_orchestration` (9): a mock encoding with countable
  circuits pins the sequencing itself — step count per `power`, adjoint
  selection, uncompute toggling, roundtrip identity, control-state
  conventions, and the QSVT phase/direction interleaving against an explicit
  matrix product.

The full `tests/python` suite (including the existing fermion/stateprep
binding tests) passes against a built extension; the new tests also run in a
source checkout with no build.

## Examples and docs

- `examples/pauli_lcu_qsvt/pauli_lcu_demo.py` — printed walkthrough of the
  encoding surface.
- `examples/pauli_lcu_qsvt/hamiltonian_simulation_qsvt.py` — QSVT time
  evolution with QSPPACK-generated phases validated against exact
  diagonalization (fidelity 1.0 to machine precision; needs `qsppack`,
  `scipy`).
- `docs/pauli_lcu_qsvt.md` — API surfaces, the hardware-shaped vs
  simulation-only split, sign conventions, and implementation notes.


python/cudaq_algorithms/__init__.py
```diff
@@ -1,5 +1,54 @@
...
+from .block_encoding import BlockEncoding
+from .common_kernels import (controlled_reflect_about_zero,
+                             controlled_signal_phase, reflect_about_zero,
+                             signal_phase)
+from .pauli_lcu import (PauliLCU, adjoint_walk, apply,
+                        apply_controlled_phase_sequence, apply_phase_sequence,
+                        controlled_adjoint_walk,
+                        controlled_reflect_about_prepare, controlled_select,
+                        controlled_walk, prepare, reflect_about_prepare,
+                        select, select_observable, state_from, unprepare, walk)
```
eliotheinrich: Some of these names are generic enough that they probably shouldn't be included in the top-level __init__.py; users may import them and end up with namespace issues. Recommend keeping them in the pauli_lcu namespace.

python/cudaq_algorithms/pauli_lcu.py
```diff
@@ -0,0 +1,726 @@
...
+
+
+@cudaq.kernel
+def unprepare(ancilla: cudaq.qview, angles: list[float]):
+    """PREPARE dagger.
+
+    Hand-written inverse of ``prepare`` (deliberate: ``cudaq.adjoint`` of
+    a kernel with this loop/branch structure is not exercised by our
+    conformance checks, so the explicit reverse is kept). Guards are
+    positive blocks, not early returns (cuda-quantum#4845).
```
eliotheinrich: This makes me a little uneasy; it seems like `cudaq.adjoint` being the inverse of the `prepare` kernel should be invariant, and hand-rolling the `unprepare` means that they can drift. Any way to enforce a correctly-invertible `prepare` kernel before getting to this point?

python/cudaq_algorithms/pauli_lcu.py
```diff
@@ -0,0 +1,726 @@
...
+        pairs = [(float(c), str(w)) for w, c in hamiltonian.items()]
+    elif isinstance(hamiltonian, cudaq.SpinOperator):
+        width = num_qubits if num_qubits is not None else int(
+            hamiltonian.qubit_count)
+        pairs = []
+        for term in hamiltonian:
+            coeff = term.evaluate_coefficient()
+            if abs(coeff.imag) > 1e-10:
+                raise ValueError(
+                    "complex Hamiltonian coefficients are not supported")
```
eliotheinrich: Depending on the type of `hamiltonian` here, complex-valued coefficients throw different errors. If `hamiltonian` is a `Mapping`, (I believe) it throws an opaque `TypeError`, but if it's a `cudaq.SpinOperator`, it throws a more descriptive `ValueError`. The coefficient validation should probably be unified.

python/cudaq_algorithms/pauli_lcu.py
```diff
@@ -0,0 +1,726 @@
...
+                 include_identity: bool = True,
+                 coefficient_threshold: float = 1e-12) -> None:
+        pairs, width = _terms_from_input(hamiltonian, num_qubits)
+
+        self._num_system = width
+        self._constant = 0.0
+        kept = []
+        for coeff, word in pairs:
+            if abs(coeff) < coefficient_threshold:
+                continue
```
eliotheinrich, line 499: `_prepare_angles` guard couples two unrelated 1e-12 thresholds.

The keep-filter drops terms on `|coeff| < 1e-12`, but the guard zeroes a rotation angle on `subtree_total < 1e-12` where the total is a probability `(|c|/alpha)`. Its comment claims the branch only fires for power-of-two padding — but two real sibling terms whose combined `|c|/alpha` falls under `1e-12` also trip it, forcing their split angle to `0`. That routes all amplitude to the left sibling and zeroes the right one, so a retained term (still in `enc.terms/num_terms`) gets prepared amplitude 0 instead of `√(|c|/alpha)` — the circuit silently encodes a different operator than advertised.

The guard only fires when the whole subtree carries `< 1e-12` probability, so any dropped term had `|c|/alpha < 1e-12` and the encoded `H/alpha` shifts by `<1e-12` in norm, so this is definitely a nitpick rather than a real bug, but still suggest gating based on the index (padded zeros all lie at the end of `probs` due to L498-L499) rather than the coefficient. 


Repro:
```
import math
def prepare_angles(probs): # == _prepare_angles
    angles = []
    for layer in range(len(probs).bit_length() - 1):
        step = len(probs) >> (layer + 1)
        for node in range(1 << layer):
            s = node * step * 2
            total = sum(probs[s:s + 2 * step])
            angles.append(0.0 if total < 1e-12 else
                          2 * math.asin(math.sqrt(sum(probs[s+step:s+2*step]) / total)))
    return angles

kept = [(2e-12, "XI"), (2e-12, "IX"), (1e6, "ZZ")]   # all clear the 1e-12 coeff filter
alpha = sum(abs(c) for c, _ in kept)
probs = [abs(c) / alpha for c, _ in kept] + [0.0]     # {XI,IX} subtree total = 4e-18 < 1e-12
print("angles:", ["%.3g" % a for a in prepare_angles(probs)])  # -> ['3.14', '0', '0']
print("IX branch amp: intended %.2e, encoded 0.0" % math.sqrt(probs[1]))  # -> 1.41e-09
```

eliotheinrich (commented):
Looks great, Scott! This is a nice implementation, and leaves room for future block encoding protocols. I left a few relatively minor comments, but overall looks great.

python/cudaq_algorithms/__init__.py
wsttiger: good idea!

python/cudaq_algorithms/pauli_lcu.py
wsttiger: Agree - let me check on this!

python/cudaq_algorithms/__init__.py
wsttiger: Done in 5debc2d — the package root now exports only the object surface (`PauliLCU`, `Walk`, `QSVT`, `PhaseSequence`, `BlockEncoding`, the observables, `state_from`, and the direction constants). The composable device kernels stay in their module namespaces: `cudaq_algorithms.pauli_lcu.{prepare, select, apply, walk, apply_phase_sequence, ...}` and `cudaq_algorithms.common_kernels.{reflect_about_zero, signal_phase, ...}`. Docs, tests, and the demo import from the module namespaces now.

python/cudaq_algorithms/pauli_lcu.py
wsttiger: Checked: `cudaq.adjoint(prepare, ...)` fails at runtime on this kernel's loop/branch structure — `could not autogenerate the adjoint of a kernel` (CUDA-Q 0.15) — so adjoint autogeneration isn't available here and the hand-written reverse has to stay. To enforce the invariant you're after, 5debc2d adds `test_unprepare_inverts_prepare`, which pins prepare-then-unprepare as the identity on an arbitrary state (not just |0...0>), and `unprepare`'s docstring now records the concrete adjoint error so the duplication reads as deliberate rather than accidental. If upstream adjoint support improves, the test makes the swap safe.

python/cudaq_algorithms/pauli_lcu.py
wsttiger: Unified in 5debc2d: every input form now routes through one `_real_coefficient` validator, so a complex coefficient raises the same `ValueError` whether it arrives in a mapping, a pair list, a `SpinOperator`, or a single `SpinOperatorTerm` product (which is now accepted and normalized to a one-term operator — it previously fell into the pair-iterable branch with its own confusing error). `test_complex_coefficients_rejected_uniformly` covers all four forms.

python/cudaq_algorithms/pauli_lcu.py
wsttiger: Nice catch, and thanks for the repro. Fixed in 5debc2d by gating on `total == 0.0` instead of a threshold: padded leaves are exactly zero, while any subtree containing a retained term has strictly positive probability (the coefficient filter guarantees it), so the exact-zero test fires only for power-of-two padding — equivalent to your index-based suggestion without threading the leaf count through. Your repro is now the regression test `test_prepare_angles_keep_tiny_sibling_terms`, including the check that padded subtrees still get zero rotations.

python/cudaq_algorithms/pauli_lcu.py
wsttiger: Follow-up: dug into whether the adjoint failure is legitimate. First verified `prepare` is genuinely unitary (reconstructed its full matrix column-by-column; U·U† = I to 4e-16 — it's only `ry`/`ry.ctrl`/`x`, no measurement or reset), so the failure is an upstream limitation, and bisecting it turned up two distinct upstream bugs, now filed:

- [cuda-quantum#4898](https://github.com/NVIDIA/cuda-quantum/issues/4898) — the loud failure `prepare` hits: adjoint autogeneration rejects unitary kernels with conditionally-conjugated rotations in nested loops (the flip/rotate/unflip idiom), even though each construct adjoints fine in isolation.
- [cuda-quantum#4897](https://github.com/NVIDIA/cuda-quantum/issues/4897) — worse, found along the way: for kernels adjoint *does* accept, loop-carried classical updates (`index += 1`) are not replayed — every adjoint iteration reuses the initial value, producing a silently wrong circuit (verified to 1e-14 against the stuck-counter hypothesis).

So even where autogeneration nominally works, #4897 means we wouldn't trust it for this kernel shape yet. The hand-written `unprepare` stays, e130961 cites both issues in the docstring and the inverse-property test, and if upstream fixes land the test makes the swap to `cudaq.adjoint` safe.

eliotheinrich (approved):
Looks good to me!

PR #13, opened 2026-07-12
Add the Suzuki-Trotter Hamiltonian simulation primitive

wsttiger:
## Summary

This PR adds **Suzuki-Trotter product formulas** as a peer primitive to the
LCU/qubitization/QSVT stack from #12: term extraction and ordering,
resource estimation, kernel factories, and the composable `apply_trotter`
device kernel, supporting first-, second- (default), and fourth-order
(Forest-Ruth) formulas.

## The primitive

```python
from cudaq_algorithms import Trotter

evolution = Trotter(hamiltonian)     # dict, SpinOperator/term, or
                                     # (coeff, word) pairs — extracted,
                                     # validated, and ordered once

kernel = evolution.kernel(time=0.8, steps=4, order=2)
# -> zero-argument @cudaq.kernel: prepares |0...0> and evolves it
# (pass directly to cudaq.sample / cudaq.observe / cudaq.get_state)

state_kernel = evolution.state_kernel(time=0.8, steps=4)
# -> @cudaq.kernel(state): evolves an arbitrary initial cudaq.State
# (the input-loading path sim_utils.evolve uses)

prepared = evolution.kernel(time=0.8, steps=4, state_prep=my_prep)
# -> zero-argument circuit composing a (qubits: qview) state-prep
#    kernel with the evolution: fully hardware-shaped, no statevector

estimate = evolution.resources(steps=4, order=2)  # TrotterResourceEstimate
evolution.words, evolution.coefficients, evolution.identity_coefficient
```

Same construction idiom as the other primitives — the object holds the
Hamiltonian, the factory calls hold the evolution choices — so
`Trotter(H).kernel(time, steps, order)` reads exactly like
`Walk(enc).kernel(power)` and `QSVT(enc).kernel(sequence)`.

- **Orders 1 / 2 / 4:** first-order sweeps, symmetric (Strang) splitting,
  and Forest-Ruth (three symmetric sub-steps with time fractions
  `w1 = 1/(2 - 2^(1/3))`, `w0 = 1 - 2*w1`, precomputed constants since
  kernels cannot evaluate cube roots).
- **Identity terms:** the circuit implements the product formula for the
  non-identity part only; `identity_coefficient` reports the omitted
  global phase (a real relative phase for controlled/interference use),
  and the simulation helper reintroduces it.
- **Term ordering:** `TrotterOrdering.PRESERVE_INPUT` (default) or
  `COEFFICIENT_MAGNITUDE_DESCENDING` (stable sort).
- **Resources:** `resources(steps, order)` — both parameters required, so
  an estimate can never silently describe a different circuit than the
  kernel that was built. Rotation counts and a CNOT-decomposition proxy,
  with no execution.
- **Composition escape hatch:** `trotter.apply_trotter(coeffs, words, t,
  steps, order, qubits)` composes inside user kernels (module namespace,
  matching the export policy from #12).
- **State-preparation injection:** `kernel(..., state_prep=my_prep)`
  composes a `(qubits: cudaq.qview)` preparation kernel with the
  evolution into one zero-argument circuit (the same parameter exists on
  the LCU/qubitization/QSVT factories).
- **Simulation helper:** `sim_utils.evolve(evolution, ket, time, steps,
  order)` delegates to `state_kernel` — one home for validation and
  marshaling — and reintroduces the identity phase.

## Testing

**27 Trotter tests** (full suite across the package: 76; with the compiled
extension: 96 + 1 skip). Coverage:

- Correctness against two independent references: exact matrix
  exponentials via diagonalization (using the shared dense-reference
  helpers), and an explicit per-order Pauli-rotation stepper.
- Per-order error thresholds and asymptotic error-scaling slope fits
  (order-p error ~ dt^p for p = 1, 2, 4).
- Exactness for commuting Hamiltonians; a 14-term 4-qubit chemistry-style
  Hamiltonian; kernel interop with flattened arguments composed in user
  kernels.
- Identity/global-phase handling through both the kernel and
  `sim_utils.evolve` paths.
- Every accepted Hamiltonian input form plus rejection tests for every
  validation rule above (including regression tests for a degenerate
  empty-Hamiltonian crash and a silently-unvalidated simulation path
  caught during pre-review hardening).

## Examples and docs

- `examples/hamiltonian_simulation/trotter_chemistry.py` — chemistry-style
  Hamiltonian evolved two ways (the one-call simulation helper, and
  `apply_trotter` composed in a user kernel with state preparation),
  validated against the exact matrix exponential.
- `docs/trotter.md` — `Trotter` usage, order semantics, identity-phase
  behavior, composition, the simulation/hardware split, and one
  deliberately deferred circuit-level optimization (merging back-to-back
  half-rotations at sweep boundaries) documented as future work.

### DIFF
--- a/python/cudaq_algorithms/trotter.py
+++ b/python/cudaq_algorithms/trotter.py
@@ -362,8 +362,11 @@ class TrotterPlan:
         """Number of retained non-identity terms."""
         return len(self.coefficients)
 
-    def kernel(self):
+    def kernel(self) -> Any:
         """Return a ``@cudaq.kernel()`` applying this plan's evolution.
+
+        (``Any`` because CUDA-Q exposes no stable public Python type for
+        compiled kernel objects.)
 
         The kernel allocates ``num_qubits`` qubits in |0...0> and applies
         the product formula. The identity phase is not included (it cannot
--- a/python/cudaq_algorithms/trotter.py
+++ b/python/cudaq_algorithms/trotter.py
@@ -58,7 +58,6 @@ FOREST_RUTH_W0: float = -1.7024143839193153
 #: ``(coefficient, word)`` pairs.
 HamiltonianLike = Union[Mapping[str, complex], Iterable[tuple], Any]
 
-
 # ============================================================================
 # Device kernel (module level, composable from user kernels)
 # ============================================================================
@@ -121,7 +120,6 @@ def apply_trotter(coefficients: list[float], words: list[cudaq.pauli_word],
 
 # state_from lives with the block-encoding module; re-exported here since
 # Trotter workflows need the same precision-aware state construction.
-
 
 # ============================================================================
 # Host-side term extraction
@@ -213,8 +211,8 @@ def _word_pairs_from_input(
 
 
 def make_trotter_terms(
-        hamiltonian: HamiltonianLike,
-        coefficient_tolerance: float = 1e-12
+    hamiltonian: HamiltonianLike,
+    coefficient_tolerance: float = 1e-12
 ) -> tuple[list[float], list[str], float, int]:
     """Return flattened terms for the Suzuki-Trotter circuit primitive.
 
@@ -295,9 +293,8 @@ def _coerce_ordering(ordering: TrotterOrdering | str) -> TrotterOrdering:
         raise ValueError(f"unsupported Trotter ordering: {ordering}") from exc
 
 
-def _ordered_terms(
-        coefficients: list[float], words: list[str],
-        ordering: TrotterOrdering) -> tuple[list[float], list[str]]:
+def _ordered_terms(coefficients: list[float], words: list[str],
+                   ordering: TrotterOrdering) -> tuple[list[float], list[str]]:
     """Apply the ordering strategy to parallel coefficient/word lists."""
     coefficients = list(coefficients)
     words = list(words)
@@ -355,8 +352,8 @@ class Trotter:
 
     def __init__(self,
                  hamiltonian: HamiltonianLike,
-                 ordering: TrotterOrdering | str = (
-                     TrotterOrdering.PRESERVE_INPUT),
+                 ordering: TrotterOrdering
+                 | str = (TrotterOrdering.PRESERVE_INPUT),
                  *,
                  coefficient_tolerance: float = 1e-12) -> None:
         ordering = _coerce_ordering(ordering)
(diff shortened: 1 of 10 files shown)


python/cudaq_algorithms/trotter.py
```diff
@@ -0,0 +1,556 @@
...
+# precomputed as module constants because kernels cannot compute cube
+# roots (host-only math).
+_FOREST_RUTH_W1: float = 1.3512071919596578
+_FOREST_RUTH_W0: float = -1.7024143839193153
+
+#: Accepted Hamiltonian input forms: a ``cudaq.SpinOperator`` (or a single
+#: ``cudaq.SpinOperatorTerm`` product), a ``{"XZI...": coefficient}``
+#: mapping, or an iterable of ``(coefficient, word)`` pairs.
+HamiltonianLike = Union[cudaq.SpinOperator, cudaq.SpinOperatorTerm,
+                        Mapping[str, complex], Iterable[tuple]]
```
eliotheinrich: `Mapping[str, complex]` should be `Mapping[str, float]`, I think

python/cudaq_algorithms/trotter.py
```diff
@@ -0,0 +1,556 @@
...
+            time, steps, order)
+        num_qubits = self._num_qubits
+
+        if state_prep is not None:
+            if not words:
+
+                @cudaq.kernel
+                def prep_identity():
+                    qubits = cudaq.qvector(num_qubits)
+                    state_prep(qubits)
```
eliotheinrich, line 439: There are a few spots that exercise special logic for when `words` is empty, i.e. for an identity Hamiltonian. I don't see any test coverage for this case; I'd like if we could add one so that these paths get executed

eliotheinrich (approved):
Other than that, looks good to me!

python/cudaq_algorithms/trotter.py
wsttiger: The `complex` is deliberate: chemistry pipelines routinely hand over coefficients as `0.5+0j`-style complex values, so the extraction accepts complex input and validates that the imaginary part is negligible (|imag| <= the same 1e-10 noise tolerance `PauliLCU` uses; larger imaginary parts raise). There's a test pinning that behavior (`test_complex_noise_tolerance_matches_pauli_lcu` / `test_make_trotter_terms_validation`), so the annotation reflects what the function actually accepts. Happy to switch the hint to `float` if you'd rather advertise the strict contract and treat complex acceptance as an implementation detail — your call.

python/cudaq_algorithms/trotter.py
wsttiger: Added in 6f2c1a8: the new `test_identity_only_hamiltonian_kernel_factory_branches` executes both empty-words branches of `kernel()` — without `state_prep` the circuit is the identity on |0...0>, and with `state_prep` it is exactly the preparation — and samples both forms as well, since the `sample` launcher is a different marshaling path than `get_state` (these branches exist precisely because of the captured-empty-list marshaling issue, cuda-quantum#4847, so they now have a regression guard). The `state_kernel` empty branch was already covered via `test_identity_only_hamiltonian_is_a_global_phase`.

python/cudaq_algorithms/trotter.py
eliotheinrich: Gotcha; makes sense. In that case, should https://github.com/NVIDIA/cudaq-algorithms/blob/4f1dbf6d48bf70124e8d3030f14ed210af5eab71/python/cudaq_algorithms/pauli_lcu.py#L52-L53 be changed for consistency?

wsttiger: Good point — done in 4b937ed: `pauli_lcu.HamiltonianLike` now advertises the same contract (`Mapping[str, complex]` / `Iterable[tuple[complex, str]]`); both extraction paths were already sharing `_real_coefficient`, so this just makes the annotations match the shared behavior. Trotter's pair form is parameterized the same way for full symmetry.

PR #14, opened 2026-07-12
Add state-preparation injection to the primitive kernel factories

wsttiger:
## Summary

Every kernel factory in the LCU/qubitization/QSVT stack now accepts an
optional **`state_prep`** kernel — signature `(qubits: cudaq.qview)` —
composing input-state preparation directly into the returned circuit.
With it, factories return **zero-argument, fully hardware-shaped
circuits**: the system register is allocated in |0...0>, `state_prep`
runs on it, and the operation follows. No `cudaq.State` data anywhere —
directly sampleable and synthesizable.

Without `state_prep`, nothing changes: the factories keep returning the
`cudaq.State`-taking circuits (the simulation-friendly form used by
`sim_utils` and the dense-reference tests).

## Motivation

The `State`-taking factories are ideal for simulation but not executable
on hardware — a QPU takes circuits, not statevectors. Until now the
hardware path required hand-composing the module-level kernels inside a
user kernel. This PR makes the factory ergonomics cover the hardware
path too, and it is the seam the upcoming state-preparation primitives
(and a QPE example) plug into.

## Contract and mechanism

- `state_prep` acts only on the register it is handed, which arrives in
  |0...0> with width `num_system`. The width cannot be verified at
  factory time (documented contract); a mismatched prep fails at launch.
- The composition mechanism is the same closure-captured kernel
  injection the `BlockEncoding` protocol already relies on — no new
  kernel-language surface.

## Testing

7 tests (`test_state_prep_injection.py`); full package suite 56, and
76 + 1 skip with the compiled extension. Every prep-mode circuit is
pinned against its `State`-taking twin fed the identical prepared state
as data: encode, walks (forward/adjoint, with and without uncompute),
roundtrips, controlled variants at both control states, QSVT plain and
controlled, and the moment path (both entry modes agree to 1e-10 across
orders). The exactly-one-input-mode validation is covered, and a
composed zero-argument circuit is sampled directly.

## Related

- The Suzuki-Trotter PR adds the same parameter to `Trotter.kernel`.
- The state-preparation primitives (Givens-rotation Slater-determinant
  prep, Hartree-Fock + fixed-parameter UCC) will ship as factories
  producing exactly this `(qubits: cudaq.qview)` kernel shape.

### DIFF
--- a/python/cudaq_algorithms/qsvt.py
+++ b/python/cudaq_algorithms/qsvt.py
@@ -190,11 +190,14 @@ class QSVT:
                sequence: PhaseSequence | Iterable[float],
                convention: str | None = None,
                state_prep: Kernel | None = None) -> Kernel:
-        """A ``@cudaq.kernel(state)`` applying the phase/walk sequence.
+        """A kernel applying the phase/walk sequence.
 
         ``sequence`` may be a PhaseSequence or a plain list of phases
-        (optionally with ``convention="qsp"``). The signal register is
-        allocated in |0...0> after the system register.
+        (optionally with ``convention="qsp"``). Without ``state_prep``
+        the returned kernel takes one ``cudaq.State`` argument; with
+        ``state_prep`` — a ``(qubits: cudaq.qview)`` kernel — it takes
+        no arguments and prepares the system register itself. The signal
+        register is allocated in |0...0> after the system register.
         """
         seq = _as_sequence(sequence, convention)
         phases = seq.projector_phases
@@ -245,12 +248,15 @@ class QSVT:
                           convention: str | None = None,
                           control_state: int = 1,
                           state_prep: Kernel | None = None) -> Kernel:
-        """``@cudaq.kernel(state)`` applying the sequence controlled.
+        """A kernel applying the sequence controlled.
 
-        Allocates the system register from ``state``, then one register
-        holding [control, signal] (a CUDA-Q Python control set cannot mix a
-        bare qubit with a separate register). With control |0> the sequence
-        is the identity.
+        Input modes as in ``kernel``: a ``cudaq.State``-taking kernel, or
+        a zero-argument kernel when ``state_prep`` is given (the injected
+        prep runs on the system register only, uncontrolled). Either way
+        the system register is followed by one register holding
+        [control, signal] (a CUDA-Q Python control set cannot mix a bare
+        qubit with a separate register). With control |0> the sequence is
+        the identity.
         """
         seq = _as_sequence(sequence, convention)
         phases = seq.projector_phases
--- a/python/cudaq_algorithms/qsvt.py
+++ b/python/cudaq_algorithms/qsvt.py
@@ -188,7 +188,8 @@ class QSVT:
 
     def kernel(self,
                sequence: PhaseSequence | Iterable[float],
-               convention: str | None = None) -> Kernel:
+               convention: str | None = None,
+               state_prep: Kernel | None = None) -> Kernel:
         """A ``@cudaq.kernel(state)`` applying the phase/walk sequence.
 
         ``sequence`` may be a PhaseSequence or a plain list of phases
@@ -201,7 +202,27 @@ class QSVT:
         # empty list captures cannot be marshaled.
         directions = list(seq.walk_directions) or [FORWARD]
         n_anc = self._encoding.num_ancilla
+        n_sys = self._encoding.num_system
         u_a = self._encoding_kernel("apply_kernel")
+
+        if state_prep is not None:
+
+            @cudaq.kernel
+            def prep_qsvt_kernel():
+                system = cudaq.qvector(n_sys)
+                state_prep(system)
+                signal = cudaq.qvector(n_anc)
+                signal_phase(signal, phases[0])
+                for i in range(1, len(phases)):
+                    if directions[i - 1] == 1:
+                        reflect_about_zero(signal)
+                        u_a(signal, system)
+                    else:
+                        u_a(signal, system)
+                        reflect_about_zero(signal)
+                    signal_phase(signal, phases[i])
+
+            return prep_qsvt_kernel
 
         @cudaq.kernel
         def qsvt_kernel(state: cudaq.State):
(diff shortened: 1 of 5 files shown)


python/cudaq_algorithms/qubitization.py
```diff
@@ -245,36 +370,60 @@ def controlled_roundtrip(state: cudaq.State):
...
         if int(count) != count or count < 0:
             raise ValueError("count must be a non-negative integer")
+        if state_prep is not None:
+            if ket is not None:
+                raise ValueError("provide exactly one of ket or state_prep")
+            return [
+                self.moment(None, order, state_prep=state_prep)
+                for order in range(int(count))
+            ]
         state = ket if isinstance(ket, cudaq.State) else state_from(ket)
```
eliotheinrich, line 435: Calling `moments(None, ..., state_prep=None)` ends up calling `state_from(None)` on this line, which probably does something nasty. The guard on L386 of this file should protect you from that scenario. Might also want to add test coverage to `test_moment_requires_exactly_one_input_mode` for this case.

tests/python/test_state_prep_injection.py
```diff
@@ -0,0 +1,130 @@
...
+from cudaq_algorithms import sim_utils as sim
+
+HAMILTONIAN = {"ZI": 0.70, "IZ": -0.43, "XX": 0.19, "YZ": 0.11}
+THETA0, THETA1 = 0.37, -0.52
+
+
+@cudaq.kernel
+def product_prep(qubits: cudaq.qview):
+    rx(0.37, qubits[0])
+    ry(-0.52, qubits[1])
```
eliotheinrich: I think these magic numbers are supposed to be THETA0 and THETA1 defined above?

python/cudaq_algorithms/qubitization.py
```diff
@@ -245,36 +370,60 @@ def controlled_roundtrip(state: cudaq.State):
...
         if int(order) != order or order < 0:
             raise ValueError("order must be a non-negative integer")
         order = int(order)
         power = order // 2
+        prep_kwargs = {"state_prep": state_prep}
         if order % 2 == 0:
             # Geometry-only (2|0..0><0..0| - I on the ancillas): derivable
             # for any zero-flagged encoding, no encoding hook needed.
-            kernel = self.kernel(power=power, uncompute=True)
+            kernel = self.kernel(power=power, uncompute=True, **prep_kwargs)
```
eliotheinrich: These prep_kwargs aren't doing anything, can just pass `state_prep=state_prep` as a kwarg to `self.kernel`, unless there's some reason to expect to pass more kwargs later

tests/python/test_state_prep_injection.py
```diff
@@ -0,0 +1,130 @@
...
+    with pytest.raises(ValueError, match="exactly one"):
+        walk.moment(None, 1)
+    with pytest.raises(ValueError, match="exactly one"):
+        walk.moments(ket, 3, state_prep=product_prep)
+
+
+def test_prep_mode_returns_zero_argument_kernels():
+    # The injected form must be directly sampleable: no arguments at all.
+    enc = PauliLCU(HAMILTONIAN)
+    kernel = Walk(enc).kernel(power=1, state_prep=product_prep)
```
eliotheinrich: Can you test the other kernels made arg-less in this PR are able to run as expected here?

python/cudaq_algorithms/qubitization.py
eliotheinrich: `walk_prepared` and `prep_walk` semantically represent the same thing in different branches here; should they have the same name?

python/cudaq_algorithms/qsvt.py
```diff
@@ -188,7 +188,8 @@ def __repr__(self) -> str:
 
     def kernel(self,
                sequence: PhaseSequence | Iterable[float],
-               convention: str | None = None) -> Kernel:
+               convention: str | None = None,
+               state_prep: Kernel | None = None) -> Kernel:
         """A ``@cudaq.kernel(state)`` applying the phase/walk sequence.
```
eliotheinrich: There are a few docstrings that are stale now that these kernels can be made argless. Here, in `QSVT.controlled_kernel`, and `Walk.controlled_kernel` all have the same semi-stale docstring issue.

eliotheinrich (commented):
I added some suggestions, otherwise looks good!

python/cudaq_algorithms/qubitization.py
wsttiger: Good catch — and it was nastier than "probably does something nasty": `state_from(None)` hits a native assertion and aborts the process rather than raising. `moments` now applies the same exactly-one-input-mode guard as `moment`, and the both-None case is covered in `test_moment_requires_exactly_one_input_mode`. Fixed in 31f8669.

tests/python/test_state_prep_injection.py
wsttiger: Yes — they were meant to be the constants. The prep kernel now captures `THETA0`/`THETA1` directly, so the kernel and the dense reference can't drift apart. Fixed in 31f8669.

python/cudaq_algorithms/qubitization.py
wsttiger: No future kwargs planned — replaced with a plain `state_prep=state_prep` pass-through. Fixed in 31f8669.

tests/python/test_state_prep_injection.py
wsttiger: Done — `test_prep_mode_returns_zero_argument_kernels` now samples every kernel this PR makes arg-less (encode/walk on the encoding, all Walk factories including the uncompute=False and controlled variants, and both QSVT factories) through the `sample` launcher, which exercises a different marshaling path than `get_state`. Also added: a no-op-prep vs |0...0>-data equivalence test and a test that two kernels minted by one factory with different preps don't cross-contaminate. In 31f8669.

python/cudaq_algorithms/qubitization.py
wsttiger: Agreed they name the same circuit. They can't literally share a name (both may be minted in one process and distinct names keep stack traces unambiguous), so the convention is now: prep-mode kernel = `prep_` + its State-taking twin's name, uniformly (`prep_walk_prepared`, `prep_controlled_walked_prepared`, matching what pauli_lcu.py and qsvt.py already did). Fixed in 31f8669.

python/cudaq_algorithms/qsvt.py
wsttiger: All three (`QSVT.kernel`, `QSVT.controlled_kernel`, `Walk.controlled_kernel`) now describe both input modes, in the same wording as `Walk.kernel`, and the controlled ones state explicitly that the injected prep runs uncontrolled (prepared once, as in QPE). Fixed in 31f8669.

eliotheinrich (approved):

PR #15, opened 2026-07-14
Convert the fermion-to-qubit transforms to pure Python

wsttiger:
## Summary

`cudaq_algorithms.fermion.jordan_wigner` and `.bravyi_kitaev` are now
implemented in pure Python with the same signatures and integral
conventions as the compiled versions they replace. 

## Design

Both transforms are one construction. A fermion-to-qubit encoding is an
invertible GF(2) matrix `A` storing occupations as qubit bits
(`x = A n mod 2`): Jordan-Wigner is the identity; Bravyi-Kitaev is the
Fenwick (binary-indexed-tree) partial-sum matrix. Three qubit masks per
mode, derived from `A` and its GF(2) inverse (update / parity / flip),
determine the ladder operators, and the Hamiltonian is compiled term by
term in a symplectic Pauli representation where a word is a pair of
bitmasks and a product is two XORs and a phase. 

```
H = scalar_offset * I + sum h[p, q] adag_p a_q
                      + sum V[p, q, r, s] adag_p adag_q a_r a_s
```

Both accept `(n, n)`, `(n, n) + (n, n, n, n)`, or `(n, n, n, n)`-alone
tensors, matching the compiled binding surface.

## Testing

New `test_fermion_compilers.py` (~20 tests), all against independent
references and (except the parity tests) requiring no compiled
extension:

- Jordan-Wigner against dense ladder-operator matrices built in NumPy,
  including non-Hermitian tensors and one-body-only / two-body-only
  call forms.
- Bravyi-Kitaev against the exact per-pair operators from the C++ unit
  tests (five Seeley-Richard-Love cases at 20 qubits, zero-error match)
  and the H2 known-answer operator.
- Three-way spectral consistency at 4-10 qubits: random Hamiltonians
  with physical electronic-structure symmetries are fully diagonalized
  as the exact Fock-space matrix (sparse ladder operators, independent
  of both transforms), the Jordan-Wigner matrix, and the Bravyi-Kitaev
  matrix — all three spectra agree to 1e-10 pairwise.
- JW/BK isospectrality on unstructured Hermitian tensors; the O(log n)
  Bravyi-Kitaev word-weight advantage.
- Validation errors, scalar offset, tolerance semantics, empty input.

The pre-existing `test_fermion.py` and `test_jordan_wigner.py` now
exercise the pure implementations and pass without the extension. Full
suite with the extension: 89 passed, 1 skipped.


ivanbasov (commented):
Elegant unification — both transforms as one GF(2) encoding (JW = identity, BK = Fenwick), and the test suite is strong (exact known-answers, three-way spectral agreement, frozen FCI). Read as correct. A few inline notes below: two test-robustness gaps, one operator-width edge case, and three low-severity API/consistency points. None block correctness.

tests/python/test_fermion_compilers.py
```diff
@@ -0,0 +1,591 @@
...
+    for n in range(dim):
+        bits = [(n >> j) & 1 for j in range(num_modes)]
+        image_bits = encoding.dot(bits) % 2
+        image = sum(int(b) << j for j, b in enumerate(image_bits))
+        perm[image, n] = 1.0
+    return perm
+
+
+@pytest.mark.parametrize("seed,m", [(11, 3), (12, 4), (13, 5), (14, 6)])
+def test_bravyi_kitaev_is_permuted_jordan_wigner(seed, m):
```
ivanbasov, line 560: The permutation and spectral BK checks pin the encoding *machinery* but are invariant to *which* invertible matrix `_fenwick_matrix` returns, so they don't independently constrain the Fenwick matrix itself. `_encoding_permutation` builds `P` from the same `_fenwick_matrix` that `bravyi_kitaev` uses, so `BK == P·JW·Pᵀ` is a self-consistency identity that holds for *any* invertible GF(2) encoding — it verifies that `_Encoding` correctly implements conjugation by `P`, not that the matrix is Fenwick. The isospectral and three-way spectral tests are basis-invariant for the same reason. The only tests that independently pin the actual encoding are the five single-pair known-answers (modes {0,1,2,6,7,18,19}) and H2 (m=4) — a sparse mode set; a single-node off-by-one at a mode in {3,4,5,8..17} would leave the whole suite green while emitting a non-standard operator. Cheap fix, either one: (a) assert `_fenwick_matrix(n)` equals an independently-constructed binary-indexed-tree matrix for several n covering mid-range modes; or (b) add one or two higher-mode single-pair known-answers from an independent BK reference.

python/cudaq_algorithms/fermion/_compilers.py
```diff
@@ -0,0 +1,331 @@
...
+            if x & bit and z & bit:
+                term = term * cudaq.spin.y(qubit)
+            elif x & bit:
+                term = term * cudaq.spin.x(qubit)
+            elif z & bit:
+                term = term * cudaq.spin.z(qubit)
+            qubit += 1
+        operators.append(coefficient * term)
+    if not operators:
+        return cudaq.SpinOperator.empty()
```
ivanbasov, line 257: The returned operator's width is implicit in the qubits actually touched — `_to_spin_operator` walks only the set bits of `x|z` and never declares a total qubit count, and this fully-pruned branch returns `cudaq.SpinOperator.empty()` with no qubit degree. Two caller-visible consequences: (a) an `(n, n)` tensor whose highest orbital has only zero / sub-tolerance couplings yields a `2^(n-1)` matrix from `to_matrix()` instead of `2^n`; (b) an all-zero / fully-pruned Hamiltonian (e.g. `jordan_wigner(np.zeros((3, 3)))`) yields a ~1×1 matrix instead of a `2^n` zero matrix. This matches the retired binding (same term-by-term construction), so it's not a regression, but it's a sharp edge worth pinning. Suggest either threading `n` into `_to_spin_operator` and seeding an n-wide identity so the register width is preserved, or documenting that the width tracks touched qubits (callers needing a fixed `n` must pass explicit dimensions) — plus a test on `to_matrix().shape` for an idle-top-orbital and a fully-pruned input.

python/cudaq_algorithms/fermion/_compilers.py
```diff
@@ -0,0 +1,331 @@
...
+    term; input entries and compiled terms with magnitude below
+    ``tolerance`` are dropped. Returns a ``cudaq.SpinOperator``.
+    """
+    one_body, two_body_arr, n = _validate_tensors(one_body_or_two_body,
+                                                  two_body)
+    return _compile_hamiltonian(_Encoding(_identity_matrix(n)), one_body,
+                                two_body_arr, scalar_offset, tolerance)
+
+
+def bravyi_kitaev(one_body_or_two_body: ArrayLike,
```
ivanbasov, line 334: Migration note worth surfacing here (and in the release notes): the retired C++ `bravyi_kitaev` antisymmetrized the two-body tensor internally (`Ṽ(p,q,r,s) = V[pqrs] − V[qprs] − V[pqsr] + V[qpsr]`, over a restricted index set), whereas this compiles each entry literally as `V[i,j,k,l]·a†_i a†_j a_k a_l`, identical to `jordan_wigner`. The new literal, JW-consistent semantics are intended and repair a prior JW/BK inconsistency — no code change needed. But a caller who previously passed a raw, non-antisymmetrized (e.g. chemist-ordered) two-body tensor and relied on that internal antisymmetrization will now silently get a different operator with no error. The module docstring documents the literal semantics; a one-line note on the `bravyi_kitaev` docstring and in the migration/release notes telling such callers to re-antisymmetrize their inputs would prevent a silent numerical surprise.

python/cudaq_algorithms/fermion/_compilers.py
```diff
@@ -0,0 +1,331 @@
...
+        matrix[i - 1, low:i] = 1
+    return matrix
+
+
+# ----------------------------------------------------------------------
+# Hamiltonian compilation
+# ----------------------------------------------------------------------
+
+
+def _validate_tensors(
```
ivanbasov, line 198: The retired binding validated shape/rank in C++ with `throw std::runtime_error(...)`, which nanobind surfaces as Python `RuntimeError`; `_validate_tensors` now raises `ValueError` for all of these cases. `ValueError` is the more Pythonic choice and I'd lean toward keeping it — but it's a silent behavioral change: a caller upgrading from the compiled extension with `try: ... except RuntimeError:` will no longer catch the shape error. Either keep `ValueError` and note it as a breaking change in the migration/release notes, or raise `RuntimeError` to preserve compatibility.

python/cudaq_algorithms/fermion/_compilers.py
```diff
@@ -0,0 +1,331 @@
...
+        operators = paired
+    return operators[0].canonicalize()
+
+
+def _compile_hamiltonian(encoding: _Encoding, one_body: np.ndarray,
+                         two_body: np.ndarray, scalar_offset: float,
+                         tolerance: float):
+
+    def negligible(value: complex) -> bool:
+        return abs(value.real) < tolerance and abs(value.imag) < tolerance
```
ivanbasov: `negligible()` prunes input entries componentwise (`abs(value.real) < tolerance and abs(value.imag) < tolerance`), but both this module's docstrings ("entries ... with magnitude below `tolerance` are dropped") and the output-side prune in `_to_spin_operator` (`abs(coefficient) < tolerance`) use a *magnitude* cutoff. The componentwise square region is strictly larger than the magnitude disk, so an entry with magnitude in `[tolerance, √2·tolerance)` is dropped even though the contract says keep it — e.g. `one_body[0,1] = 8e-7 + 8e-7j` (|v| ≈ 1.13e-6) at `tolerance = 1e-6`. Harmless at the default `1e-15` and Hermiticity-preserving (conjugate partners drop together), but it diverges from the docs and the output filter at any coarser user tolerance. Suggest `return abs(value) < tolerance` for consistency. (It currently mirrors the retired C++ `is_complex_zero`, so this is a deliberate doc-consistency change, not just parity.)

tests/python/test_fermion_compilers.py
```diff
@@ -0,0 +1,591 @@
...
+    assert max_weight(jordan_wigner(one_body)) == m
+    assert max_weight(bravyi_kitaev(one_body)) < m // 2
+
+
+# ----------------------------------------------------------------------
+# API surface
+# ----------------------------------------------------------------------
+
+
+def test_scalar_offset_and_tolerance():
```
ivanbasov, line 395: The output-side tolerance trim is never exercised. `test_scalar_offset_and_tolerance` drops its `1e-9` entry input-side via `negligible()` before compilation, so it never reaches `abs(coefficient) < tolerance` in `_to_spin_operator` (the analog of the retired `.canonicalize().trim(tolerance)`). Because the suite's comparison tolerances (1e-12 / 1e-10 / 1e-4) are far looser than the default prune (1e-15), mutating that line to `<=` or deleting it leaves every test green — so the behavior most likely to diverge from the old binding for realistic Hamiltonians (roundoff residuals cancelling near tolerance) is unverified. Suggest a case whose *compiled* coefficient lands near the tolerance boundary, asserting the term is/isn't present. (Note: one-body entries compile to words of magnitude ~|c|/2, so to probe the input threshold specifically the entry magnitude must exceed `2·tolerance`, else the output trim prunes it even under the correct implementation.)

wsttiger:
Thanks @ivanbasov — really thorough review, much appreciated. All six addressed in 4b8a4b6 (suite green: 30 passed, 27 existing + 3 new): the two test-robustness gaps (Fenwick pin, output-side prune), the width edge (documented + pinned), the `negligible()` magnitude fix, and migration notes for the `ValueError` and BK-antisymmetrization changes. Point-by-point in the threads. One open question on the width item — happy to thread `n` through for fixed-width output if you'd prefer that over documenting the touched-qubits behavior.

tests/python/test_fermion_compilers.py
wsttiger: Good catch — agreed the `BK == P·JW·Pᵀ` and spectral tests are invariant to *which* invertible P we use, and the single-pair known-answers skip the mid-range modes. Added `test_fenwick_matrix_matches_binary_indexed_tree` in 4b8a4b6: it pins `_fenwick_matrix(n)` against an independent Fenwick formulation (the `i += i & -i` update traversal vs. the implementation's range-marking) for n = 1..20, plus a hand-written n=8 anchor. A single-node off-by-one now fails the suite.

tests/python/test_fermion_compilers.py
wsttiger: Fixed in 4b8a4b6 — added `test_output_side_tolerance_trims_small_compiled_terms`. Using your ~|c|/2 observation, at `tolerance=1e-6` it uses c=3e-6 (compiled Z coefficient 1.5e-6, kept) and c=1.5e-6 (survives the input prune but compiles to 0.75e-6, dropped by the output-side prune). Removing or mutating that prune line now fails.

python/cudaq_algorithms/fermion/_compilers.py
wsttiger: Fixed to `abs(value) < tolerance` in 4b8a4b6, matching both the docstrings and the output-side prune. Agreed it's harmless at the default `1e-15` and Hermiticity-preserving, but the componentwise square diverging from the documented magnitude disk at coarser tolerances wasn't worth keeping.

python/cudaq_algorithms/fermion/_compilers.py
wsttiger: Documented + pinned rather than changed, to keep parity with the retired binding (and because forcing a fixed `2^n` would mean injecting a width-pinning term, which has its own surprises). 4b8a4b6 adds a module-docstring note — width tracks touched qubits, so an idle top orbital or a fully-pruned Hamiltonian yields a narrower/empty operator — and `test_operator_width_tracks_touched_qubits` pins both the idle-top-orbital (`2^1`) and all-zero (`0×0`) `to_matrix()` shapes. Happy to thread `n` through for fixed-width output instead if you'd prefer that.

python/cudaq_algorithms/fermion/_compilers.py
wsttiger: Added in 4b8a4b6 — the `bravyi_kitaev` docstring now states the two-body tensor is compiled literally (identical to `jordan_wigner`), that this deliberately repairs the prior JW/BK inconsistency, and that a caller who relied on the retired binding's internal antisymmetrization must antisymmetrize their input themselves. Also summarized in the module docstring's new migration section; will carry into the release notes when we cut them.

python/cudaq_algorithms/fermion/_compilers.py
wsttiger: Kept `ValueError` (agree it's the Pythonic choice) and called it out as a behavior change in the module docstring's new "Migration from the compiled extension" note, so upgraders with `except RuntimeError` are warned; will carry into the release notes when we cut them.

python/cudaq_algorithms/fermion/_compilers.py
ivanbasov: Documenting + pinning is the right call — agreed that threading `n` through isn't worth it (a width-pinning term has its own surprises, and parity with the retired binding is worth keeping). One correction to the new bullet, though: `to_matrix()` spans the *set* of qubits actually touched, not the range up to the highest one (verified against this branch on cudaq 0.15):

- "ensure the top orbital is touched" isn't sufficient for a fixed `2^n`: `jordan_wigner(np.diag([1.0, 0.0, 1.0]))` touches the top orbital yet `to_matrix()` is 4×4, not 8×8 — an untouched *interior* orbital narrows it too.
- "a Hamiltonian that never couples the highest orbital(s) yields a narrower operator" is guaranteed for JW but not for BK: the Fenwick update masks reach ancestor qubits — at n=6 a single hopping term between orbitals 3 and 4 touches qubit 5 (touched set {1,2,3,4,5}) even though orbital 5 has no couplings.
- The fully-pruned case yields a 0×0 matrix rather than `2^0` (your own test pins this).

Suggested wording: "the operator's width tracks the set of qubits actually touched — an untouched interior qubit also narrows `to_matrix()`, and under BK the Fenwick masks may touch qubits above the highest coupled orbital; callers needing a fixed `2^n` must pad the result themselves; a fully-pruned Hamiltonian yields an empty operator (0×0 matrix)." Non-blocking — approving as is.

tests/python/test_fermion_compilers.py
ivanbasov: Confirmed — deleting the prune line is now caught (`IZ` at 0.75e-6 would appear). One residual on the "or mutating" half: the `<` → `<=` mutant still survives, since no compiled coefficient in the suite lands exactly on the tolerance boundary. If you want to pin strictness: a diagonal entry `c = 2e-6` at `tolerance = 1e-6` compiles to a Z coefficient of exactly `1e-6` (the ×0.5 is exact in floats; verified on this branch that the term is kept under `<`), and it would be dropped under `<=`. Non-blocking.

ivanbasov (approved):
All six addressed and verified — I cross-checked the new Fenwick test's reference against an independent binary-indexed-tree construction (n = 1..24, and the n=8 anchor is exact), the prune arithmetic, and the migration notes against the code; CI is green and the removed empty-input test's coverage is preserved (and strengthened) in `test_operator_width_tracks_touched_qubits`. On the width question: keep the documented touched-qubits behavior — threading `n` through isn't worth the width-pinning term it would take. Two non-blocking notes in the threads: a wording fix for the width bullet (`to_matrix()` spans the touched *set*, and the "narrower operator" claim doesn't hold for BK), and an optional exact-boundary case for the output-side prune.

PR #17, opened 2026-07-15
Convert the state preparation to pure Python and retire the native build

wsttiger:
**Branch:** `wsttiger:features/convert_stateprep_python` → base: PR #15
branch (`features/fermion_compilers_python`); retarget to `main` once
#15 merges (the branch is already rebased onto #15's current head, so
the diff collapses to exactly these five commits).

---

## Title

Convert the state preparation to pure Python and retire the native build

## Body

This converts the last compiled component — the state-preparation
library — to pure Python and removes the native build entirely. The
`uccsd`, `uccgsd`, `upccgsd`, and `ceo` device kernels become
composable `@cudaq.kernel` functions; the excitation enumeration,
operator-pool builders, and Pauli-list helpers run on the host with
`cudaq.spin` algebra; and with nothing left to compile, `lib/`,
`include/`, the nanobind bindings, and the CMake build are deleted.
The package and its wheels are now pure Python.

### Equivalence to the compiled implementation

Measured against the compiled extension before its removal:

* State vectors are **bit-identical** across 16 configurations (uccsd
  up to 16 qubits / 360 parameters including an open-shell case,
  uccgsd to 10 qubits, upccgsd to 12, ceo to 4 orbitals), identical
  seeded parameters, `qpp-cpu` fp64.
* All operator pools match the C++ **term-for-term, including term
  order** (1,354 operators compared as word/coefficient lists), and
  `get_uccsd_excitations` matches at every configuration tested.
* Performance is equal within noise at every size: runtime is
  simulator-dominated over identical circuits, and host-side pool
  construction matches to the millisecond.

### Correctness beyond the C++ comparison

* `tests/python/test_stateprep_kernels.py` validates every ansatz
  kernel against dense single-term matrix-exponential references
  (scipy `expm`, no commutation assumptions) applied to the
  Hartree-Fock determinant — including the CEO same-spin doubles at
  the smallest size where they exist (4 orbitals). Tolerances are
  precision-aware (1e-12 on fp64 targets, 5e-5 on fp32).
* The kernel/pool theta conventions inherited from the C++ (half-angle
  rz gadgets; the double-excitation circuit's sign flip for
  (p&lt;q, r&gt;s) / (p&gt;q, r&lt;s) index patterns, which the pool
  operators do not carry) are reproduced exactly and documented in the
  dense-reference test.

### Build, packaging, and CI

* The wheel configs (`pyproject.toml.cu12`/`.cu13`) switch from
  scikit-build-core to setuptools and produce `py3-none-any` wheels —
  no compiler toolchain, no auditwheel repair. A locally built wheel
  contains exactly the package modules and resolves `__version__` from
  distribution metadata.
* The CI test job no longer builds anything: it still provisions
  CUDA-Q for the `cudaq` Python runtime, then runs pytest directly
  (`scripts/test_algorithms_build.sh` is now a pure test runner);
  `scipy` joins the test requirements for the dense references. This
  also removes the job's two flakiest failure classes (apt toolchain
  installs and C++ gtest discovery), both hit on other PRs this week.
* Known follow-ups deliberately left to the CI workstream: the wheel
  workflow still provisions the CUDA-Q toolchain it no longer needs,
  and the `build-wheels` PR path filter doesn't watch the package
  sources — both pre-existing behaviors worth fixing together with the
  wheel-matrix cleanup.

### Test plan

* `pytest tests/python` on the full branch: **145 passed** — the
  entire suite is green with no compiled extension, for the first
  time including the state-preparation and operator-pool tests, which
  previously required the native build.
* `scripts/test_algorithms_build.sh` (the CI entry point): green.
* Pure wheel built from `pyproject.toml.cu12`, contents inspected,
  installed, and smoke-tested.
* Cross-checked against the compiled extension (bit-identical states,
  term-exact pools) before its removal, as described above.


tests/python/test_stateprep_kernels.py
```diff
@@ -0,0 +1,264 @@
...
+    reference = _dense_product_reference(
+        term_groups, [t * s for t, s in zip(thetas, signs)],
+        num_qubits,
+        _hf_ket(num_qubits, occupation),
+        scale=-0.5)
+    _assert_close(actual, reference)
+
+
+@pytest.mark.parametrize("num_qubits", [4, 6])
+def test_uccgsd_kernel_matches_dense_exponential(num_qubits):
```
ivanbasov, line 189: This kernel test — and its upccgsd/ceo siblings below — is self-consistent rather than an independent check of the pool *contents*: `get_uccgsd_pauli_lists` feeds the kernel the pauli-lists derived from the same `make_uccgsd_operator_pool` that builds the dense reference `term_groups`, so a count-preserving error in the pool's Pauli words, signs, or index conventions would pass on both sides. Unlike UCCSD — whose pool has an absolute known-answer (`test_operator_pools.py::test_uccsd_operator_pool_correctness`) and whose kernel computes indices inline, independent of the pool — the generalized/CEO pools are pinned only by their term *counts* (`test_operator_pools.py`). The term-for-term C++ comparison that used to pin their contents isn't in the committed suite (it needed the extension). Suggest adding a small absolute known-answer for one `uccgsd`, `upccgsd`, and `ceo` pool (e.g. n=4 / num_orbitals=2), mirroring the UCCSD one, so a content regression in the generalized/CEO operators can't ship green. (Same gap covers the `only_singles` / `only_doubles` paths, which are count-checked but never run through a kernel end-to-end.)

.github/workflows/pr_workflow.yaml
```diff
@@ -62,10 +62,6 @@ jobs:
             build-algorithms:
               - '.github/workflows/lib_algorithms.yaml'
               - '.github/workflows/pr_workflow.yaml'
-              - 'CMakeLists.txt'
-              - 'cmake/**'
-              - 'include/**'
-              - 'lib/**'
               - 'python/**'
```
ivanbasov, line 65: This `build-algorithms` filter watches `python/**`, so a package-source change does run the test job — but that job (`scripts/test_algorithms_build.sh`) runs pytest against the source tree via `PYTHONPATH`, and the `build-wheels` filter just below (lines 70-78) does *not* include `python/**`. Net effect after this PR: editing `python/cudaq_algorithms/**` never builds or imports the actual wheel in CI, so a packaging regression — a submodule missed by `[tool.setuptools.packages.find]`, broken version metadata, or an import that only resolves from the source layout — would pass all PR checks even though the wheel is now the whole deliverable. Suggest either adding `python/**` to the `build-wheels` filter, or adding a lightweight `pip install .` + `python -c "import cudaq_algorithms; print(cudaq_algorithms.__version__)"` smoke step to the test job.

pyproject.toml.cu12
```diff
@@ -17,19 +17,11 @@ dependencies = [
...
+package-dir = {"" = "python"}
 
-[tool.scikit-build.cmake.define]
-CUDAQ_ALGORITHMS_BINDINGS_PYTHON = true
-CUDAQ_ALGORITHMS_INCLUDE_TESTS = false
+[tool.setuptools.packages.find]
+where = ["python"]
+include = ["cudaq_algorithms*"]
 
 [tool.setuptools_scm]
```
ivanbasov, line 27: `[tool.setuptools_scm]` has no `fallback_version`, so building the wheel from a tree without git metadata — an unpacked sdist, a shallow/exported checkout, a release tarball — fails hard at version resolution rather than degrading. (The runtime `_resolve_version()` in `__init__.py` handles the *installed*-metadata path, but that's separate from build-time version computation.) Suggest `fallback_version = "0.0.0"` (or the current release) under `[tool.setuptools_scm]` here and in `pyproject.toml.cu13`.

tests/python/test_stateprep_kernels.py
```diff
@@ -0,0 +1,264 @@
...
+         num_qubits, num_electrons, spin)
+    signs = [1.0] * (len(singles_alpha) + len(singles_beta))
+    for p, q, r, s in doubles_mixed + doubles_alpha + doubles_beta:
+        flipped = (p < q and r > s) or (p > q and r < s)
+        signs.append(-1.0 if flipped else 1.0)
+    return signs
+
+
+@pytest.mark.parametrize("num_qubits,num_electrons,spin",
+                         [(4, 2, 0), (6, 3, 1), (8, 4, 0)])
```
ivanbasov: The only open-shell (spin>0) config here is `(6, 3, 1)`, which produces no same-spin doubles: at `num_qubits=6, num_electrons=3, spin=1` you get `n_occ_alpha=2, n_virt_alpha=1, n_occ_beta=1`, so both the alpha-doubles (`n_virt_alpha-1 = 0`) and beta-doubles (`n_occ_beta-1 = 0`) loops are empty. The same-spin double *circuit* is covered by the closed-shell cases, but the open-shell occupancy split feeding the alpha/beta double indices is never exercised end-to-end. Suggest adding a spin>0 config large enough to emit same-spin doubles (e.g. `num_qubits=10/12` with electrons giving `n_occ_alpha >= 2` and `n_virt_alpha >= 2`).

ivanbasov (commented):
Careful gate-for-gate port — the documented CUDA-Q compiler workarounds and the dense-exponential kernel tests are a good sign, and I found no correctness/transcription issues in the kernels or pools. The notes below are all test-coverage and packaging: the generalized/CEO pool contents aren't independently pinned, and no PR-CI job validates the actual wheel on a package-source change. None block correctness.

wsttiger:
Thanks @ivanbasov — spot-on again. The two self-consistency gaps and the two packaging holes are all fair. Addressed in 37b76ae: independent content pins for the generalized/CEO pools, an open-shell case that actually emits same-spin doubles, `fallback_version`, and `python/**` in the wheel filter. Point-by-point in the threads.

tests/python/test_stateprep_kernels.py
wsttiger: Agreed — the kernel tests build their dense reference from the same `make_*_operator_pool` they exercise, so a count-preserving content error would pass on both sides. Added three absolute content pins in `test_operator_pools.py` (37b76ae): `test_uccgsd_operator_pool_content_matches_fermionic_generators` and the upccgsd sibling match each pool operator (as a bijection, up to the global ±i convention) against the fermionic single/double generators built independently from dense JW ladder operators — a construction sharing no code with the pool's Pauli algebra, so a wrong word/sign/index is caught. These also run the `only_singles`/`only_doubles` pools. CEO is a deliberately non-fermionic coupled-exchange construction (single = `0.5(Y_q X_p − X_q Y_p)`, no JW string), so it has no simpler physical reference; `test_ceo_operator_pool_correctness` gives it an absolute known-answer for `num_orbitals=2`, mirroring the UCCSD one.

tests/python/test_stateprep_kernels.py
wsttiger: Fixed — you're right that `(6, 3, 1)` yields no same-spin doubles (both loops empty). Added `(10, 5, 1)` to `test_uccsd_kernel_matches_dense_exponential` in 37b76ae: it's open-shell and large enough to emit same-spin **alpha and beta** doubles (3 each), so the open-shell occupancy split feeding the alpha/beta double indices is now exercised end-to-end.

pyproject.toml.cu12
wsttiger: Fixed in both `pyproject.toml.cu12` and `.cu13` (37b76ae): `fallback_version = "0.0.0"` under `[tool.setuptools_scm]`. Verified in a git-less copy of the tree — `get_version(..., fallback_version="0.0.0")` returns `0.0.0`, where without it setuptools-scm raises the `LookupError` you described.

.github/workflows/pr_workflow.yaml
wsttiger: Good catch — the wheel is the whole deliverable now, and the `build-wheels` filter didn't include `python/**` (only `python/metapackages/**`), so a pure package-source edit would skip the wheel build+test entirely. Added `python/**` to the `build-wheels` filter in 37b76ae, so the existing build-and-import-the-wheel job now runs on any package-source change. Open to the filter-vs-smoke-step call if you'd rather keep the wheel job narrowly scoped and add a lightweight `pip install . + import` to the fast test job instead.

tests/python/test_stateprep_kernels.py
ivanbasov: The uccgsd/upccgsd pins are exactly what was needed — I mutation-tested them on this branch: an intra-operator relative sign flip fails both bijection tests cleanly, and the reference construction shares no code with the pool algebra. Two residuals, both CEO — my "num_orbitals=2" suggestion turns out to be too small for that pool:

- The same-spin-doubles branch needs four descending spatial orbitals (`i > j > k > l`), so it is empty at `num_orbitals=2`, and only one mixed double exists there — those enumerations are still pinned only by counts plus the self-consistent kernel test. Verified on this branch: changing the third same-spin pairing `(q, p, r, s)` → `(q, r, p, s)` (4 of 96 pool operators change at `num_orbitals=4`, counts preserved) and swapping the i/j roles in the mixed-double emission (a no-op at 2 orbitals, 36 == 36 operators at 4) each leave the whole suite green. Extending the known answer with a few `num_orbitals=4` operators (the first same-spin triple + one more mixed double) would close it.
- (Minor) `_assert_bijection_up_to_global_phase` calls `unmatched.remove(hit)` on a list of numpy arrays — it works today only because pool order matches reference order; a content-correct reordering crashes with `ValueError: The truth value of an array ... is ambiguous` instead of passing. One-liner: `unmatched = [g for g in unmatched if g is not hit]`.

One wording note: the matching is per-operator ±i rather than one global convention — necessarily so (the correct pool's phases are mixed: the n=4 singles all match at −i, the doubles at +i, +i, −i) — which means a whole-operator sign flip still passes. That's harmless variationally (θ absorbs it), but the module docstring's "coefficient signs match the C++ implementation exactly (the tests pin them)" holds only for intra-operator relative signs; worth softening. Non-blocking — approving.

.github/workflows/pr_workflow.yaml
ivanbasov: Verified end-to-end on this branch: with `python/**` in the filter, a package-source edit builds the real wheel from `pyproject.toml.cuX`, installs it via the metapackage, imports it, and runs the full suite against the installed package (no source-tree `PYTHONPATH`). On filter-vs-smoke: keep the filter — the wheel job exercises the actual deliverable path, which a `pip install .` smoke step can't even reproduce (there's no root `pyproject.toml`; the build generates it by copying `pyproject.toml.cuX`), and the CUDA-Q wheel cache keeps the cost bounded. Optional: `tests/**` isn't in the build-wheels filter, so a tests-only edit never runs against the installed wheel — harmless for packaging, add it only if you want belt-and-suspenders.

ivanbasov (approved):
All four addressed and verified — I ran both changed test files on this branch (23 passed), checked by execution that (10, 5, 1) emits 3 alpha + 3 beta same-spin doubles through the kernel's own inlined occupancy split, traced the wheel-filter chain end-to-end (real wheel → metapackage install → import → full pytest, no source-tree PYTHONPATH), and confirmed both pyprojects now build git-less. The uccgsd/upccgsd content pins are genuinely independent — I mutation-tested them and per-term word/sign/coefficient errors fail cleanly. One follow-up worth doing in the threads: the CEO known answer at num_orbitals=2 can't reach the same-spin-doubles branch, so that enumeration is still self-consistent-only (details + concrete cases in the thread). Non-blocking; approving.

PR #18, opened 2026-07-16
Add pure-Python double factorization of two-electron integrals

wsttiger:
**Branch:** `wsttiger:features/add_python_double_factorization` → base:
`main`. Rebased onto current main (post #13/#14); five commits. The two
follow-up branches (chemistry bridge, `DoubleFactorizedEncoding`) stack
on this one.

---

## Title

Add pure-Python double factorization of two-electron integrals

## Body

This adds `cudaq_algorithms.double_factorization`: classical chemistry
preprocessing that factorizes two-electron repulsion integrals into the
low-rank leaf-rotation/core representation consumed by double-factorized
block encodings, qubitization, and measurement schemes. It implements
the explicit (X-DF) and compressed (C-DF) variants of Cohn, Motta, and
Parrish, PRX Quantum 2, 040352 (2021), plus the RC-DF regularization of
Oumarou et al. (Quantum 8, 1371 (2024)).

### What's included

* **X-DF** — exact double factorization; the first factorization
  defaults to pivoted Cholesky (rank-revealing on the PSD ERI
  supermatrix, stops at the true rank), with a symmetric-
  eigendecomposition variant for indefinite inputs. Exact to machine
  precision at full rank.
* **C-DF / RC-DF** — fixed-leaf-count compression by least squares:
  leaf rotations parameterized as `exp(X)` over antisymmetric `X` and
  optimized with L-BFGS, symmetric cores solved in closed form each
  step (the paper's two-step scheme), warm-started from X-DF;
  `regularization=rho` adds the RC-DF L2 core penalty. An optional
  matrix-free CG inner core solve (`inner_solver="cg"`, with warm
  starting) replaces the dense least-squares solve for larger systems.
* **Helpers** — ERI reconstruction, Frobenius reconstruction error,
  the DF one-body correction `kappa`, and DF Hamiltonian one-norms in
  both the Pauli-rotation (`"lcu"`) and qubitization (`"burg"`)
  conventions.
* **GPU path** — heavy linear algebra runs on cuSOLVER/cuBLAS via CuPy
  when a GPU is present, NumPy/SciPy otherwise. `backend="auto"` is
  size-aware (CPU below the empirical crossover where kernel-launch
  latency dominates; documented per method) and every entry point
  accepts an explicit `"cupy"`/`"numpy"` override. The subpackage has
  no `cudaq` dependency at all.
* Docs (`docs/double_factorization.md`), an H2O/STO-3G example (PySCF
  kept at the example boundary per the chemistry-inputs policy), a CG
  benchmark, and a test suite covering both first factorizations,
  C-DF/RC-DF, one-norm conventions, backend selection, and input
  validation — parameterized over NumPy plus CuPy when a GPU is
  present, with PySCF/OpenFermion reference tests that self-skip.

### Conventions

Integrals are chemist-notation `(pq|rs)` over real spatial orbitals
(8-fold symmetric), as from `pyscf.ao2mo.restore("s1", ...)`; the
factorization is `(pq|rs) ≈ Σ_t U^t Z^t U^tᵀ` with orthogonal leaf
rotations (compilable into Givens fabrics) and symmetric cores.

### Review

The branch went through an independent adversarial review, which
verified the mathematics quantitatively from scratch implementations
(reconstruction exactness, truncation error exactly matching the
documented eigenvalue bound, the C-DF analytic gradient against
central differences, the CG normal equations against the dense solve,
and both one-norm formulas re-derived from Pauli algebra) and found
one real defect, fixed in the final commit: **scipy is now a declared
dependency** (the package imports this subpackage unconditionally, and
it imports scipy at module load — a clean wheel install previously
failed at import). The same commit adds two guardrails the review
recommended: input tensors that violate the real-orbital chemist
symmetries are rejected (previously silently wrong answers), and the
pivoted-Cholesky path warns when it clamps a significant negative
residual on an indefinite supermatrix (previously silent accuracy
loss; the eigendecomposition path handles such inputs).

### Test plan

* `pytest tests/python` on the branch: the double-factorization suite
  passes in full — 17 passed, 1 skipped
  (OpenFermion cross-check; not installed) — and the only remaining
  failures on this base are the
  known extension-only state-prep/fermion tests, fixed by #15's line.
* Example runs end-to-end against PySCF-generated H2O integrals.

### DIFF
--- a/python/cudaq_algorithms/__init__.py
+++ b/python/cudaq_algorithms/__init__.py
@@ -15,6 +15,9 @@ The package has two kinds of components:
 - Pure-Python modules (:mod:`.pauli_lcu`, :mod:`.qubitization`, :mod:`.qsvt`,
   :mod:`.trotter`, :mod:`.sim_utils`): quantum primitives implemented as
   CUDA-Q Python kernels. These only require the ``cudaq`` Python package.
+- Pure-Python classical preprocessing (:mod:`.double_factorization`):
+  chemistry-oriented numerics on NumPy/SciPy with optional CuPy GPU
+  acceleration; no ``cudaq`` or compiled-extension dependency.
 
 The native extension is optional: if it is not present (for example in a
 source checkout without a build), the pure-Python APIs below still import
@@ -35,6 +38,9 @@ except ModuleNotFoundError as exc:
         raise
     _NATIVE_IMPORT_ERROR = exc
     __version__ = "CUDA-Q Algorithms (compiled extension not built)"
+
+# Pure-Python classical preprocessing (no compiled-extension dependency).
+from . import double_factorization
 
 # Pure-Python quantum primitives (no compiled-extension dependency).
 from . import (block_encoding, common_kernels, pauli_lcu, qsvt, qubitization,
--- a/python/cudaq_algorithms/__init__.py
+++ b/python/cudaq_algorithms/__init__.py
@@ -17,7 +17,9 @@ The package has two kinds of components:
   CUDA-Q Python kernels. These only require the ``cudaq`` Python package.
 - Pure-Python classical preprocessing (:mod:`.double_factorization`):
   chemistry-oriented numerics on NumPy/SciPy with optional CuPy GPU
-  acceleration; no ``cudaq`` or compiled-extension dependency.
+  acceleration. The subpackage itself imports neither ``cudaq`` nor the
+  compiled extension, but it is reached through this package, whose
+  import does load ``cudaq`` alongside it.
 
 The native extension is optional: if it is not present (for example in a
 source checkout without a build), the pure-Python APIs below still import
--- a/python/cudaq_algorithms/double_factorization/_backend.py
+++ b/python/cudaq_algorithms/double_factorization/_backend.py
@@ -38,7 +38,7 @@ except Exception:  # pragma: no cover - environment without CuPy
 # than NumPy; above them the batched cuSOLVER/cuBLAS linear algebra wins and the
 # lead grows with ``n``. X-DF (a cheap Cholesky loop + tiny per-leaf eigh) stays
 # CPU-favorable far longer than the contraction-heavy C-DF optimization. See
-# benchmarks/double_factorization/ and DOUBLE_FACTORIZATION_CPP_PORT_NOTES.md.
+# benchmarks/double_factorization/.
 AUTO_GPU_MIN_ORBITALS_COMPRESSED = 18
 AUTO_GPU_MIN_ORBITALS_EXPLICIT = 56
 
--- a/python/cudaq_algorithms/double_factorization/_backend.py
+++ b/python/cudaq_algorithms/double_factorization/_backend.py
@@ -97,7 +97,9 @@ def to_numpy(array) -> np.ndarray:
 def expm_skew_symmetric(generator: ArrayLike, xp: ArrayModule) -> DeviceArray:
     """Matrix exponential of a real antisymmetric matrix, returning an orthogonal
     matrix. Uses a Hermitian eigendecomposition (cuSOLVER on the CuPy backend),
-    avoiding any dependence on a general matrix-exponential routine."""
+    avoiding any dependence on a general matrix-exponential routine. Also the
+    per-leaf fallback of :func:`expm_skew_symmetric_batched` where CuPy's
+    batched eigensolver does not apply (matrix dimension > 32)."""
     generator = xp.asarray(generator)
     # i * X is Hermitian for real antisymmetric X, so eigh applies.
     hermitian = 1j * generator
@@ -112,8 +114,19 @@ def expm_skew_symmetric_batched(generators: ArrayLike,
     """Batched :func:`expm_skew_symmetric` over a stack of real antisymmetric
     matrices ``generators`` (shape ``(num_leaves, n, n)``). One batched Hermitian
     eigendecomposition instead of ``num_leaves`` separate ones -- on the CuPy
-    backend this is a single cuSOLVER call rather than a Python-driven loop."""
+    backend this is a single cuSOLVER call rather than a Python-driven loop.
+
+    CuPy dispatches stacked ``eigh`` to cuSOLVER's batched Jacobi solver
+    (``syevj``/``heevjBatched``), which is documented for matrix dimensions
+    up to 32; above that this falls back to per-leaf (non-batched)
+    eigendecompositions, which remain on the GPU through the standard
+    cuSOLVER path."""
     generators = xp.asarray(generators)
+    if xp.__name__ == "cupy" and generators.shape[-1] > 32:
+        return xp.stack([
+            expm_skew_symmetric(generators[index], xp)
+            for index in range(generators.shape[0])
+        ])
     hermitian = 1j * generators
     eigenvalues, eigenvectors = xp.linalg.eigh(
         hermitian)  # batched over axis 0
--- a/python/cudaq_algorithms/double_factorization/_backend.py
+++ b/python/cudaq_algorithms/double_factorization/_backend.py
@@ -14,7 +14,19 @@ APIs accept a ``backend`` argument of ``"auto"`` (default), ``"cupy"``, or
 """
 from __future__ import annotations
 
+from typing import Any, TypeAlias
+
 import numpy as np
+from numpy.typing import ArrayLike
+
+# The array namespace used for a computation: the ``numpy`` module or, when
+# a GPU is available, ``cupy``. CuPy is optional, so there is no importable
+# static type covering both namespaces.
+ArrayModule: TypeAlias = Any
+
+# An array living on the active backend: ``numpy.ndarray`` on the NumPy
+# backend, ``cupy.ndarray`` on the CuPy backend.
+DeviceArray: TypeAlias = Any
 
 try:  # CuPy is optional; absence simply forces the NumPy path.
     import cupy as _cupy  # type: ignore
@@ -41,7 +53,9 @@ def cupy_gpu_available() -> bool:
         return False
 
 
-def resolve_backend(backend: str = "auto", problem_size=None, gpu_min_size=0):
+def resolve_backend(backend: str = "auto",
+                    problem_size: int | None = None,
+                    gpu_min_size: int = 0) -> tuple[ArrayModule, str]:
     """Return ``(array_module, name)`` for the requested backend.
 
     ``"auto"`` selects CuPy when a GPU is available *and* the problem is large
@@ -68,7 +82,7 @@ def resolve_backend(backend: str = "auto", problem_size=None, gpu_min_size=0):
         "'auto', 'cupy', or 'numpy'.")
 
 
-def to_device(array, xp):
+def to_device(array: ArrayLike, xp: ArrayModule) -> DeviceArray:
     """Move a host/device array onto the backend ``xp``."""
     if xp is np:
         return np.asarray(array)
@@ -82,7 +96,8 @@ def to_numpy(array) -> np.ndarray:
     return np.asarray(array)
 
 
-def expm_skew_symmetric(generator, xp):
+def expm_skew_symmetric(generator: ArrayLike,
+                        xp: ArrayModule) -> DeviceArray:
     """Matrix exponential of a real antisymmetric matrix, returning an orthogonal
     matrix. Uses a Hermitian eigendecomposition (cuSOLVER on the CuPy backend),
     avoiding any dependence on a general matrix-exponential routine."""
@@ -95,7 +110,8 @@ def expm_skew_symmetric(generator, xp):
     return xp.real(rotated)
 
 
-def expm_skew_symmetric_batched(generators, xp):
+def expm_skew_symmetric_batched(generators: ArrayLike,
+                                xp: ArrayModule) -> DeviceArray:
     """Batched :func:`expm_skew_symmetric` over a stack of real antisymmetric
     matrices ``generators`` (shape ``(num_leaves, n, n)``). One batched Hermitian
     eigendecomposition instead of ``num_leaves`` separate ones -- on the CuPy
--- a/python/cudaq_algorithms/double_factorization/_backend.py
+++ b/python/cudaq_algorithms/double_factorization/_backend.py
@@ -68,18 +68,16 @@ def resolve_backend(backend: str = "auto",
         return np, "numpy"
     if backend == "cupy":
         if not cupy_gpu_available():
-            raise RuntimeError(
-                "double_factorization error - the 'cupy' backend was requested "
-                "but no CuPy/GPU is available.")
+            raise RuntimeError("the 'cupy' backend was requested but no "
+                               "CuPy/GPU is available.")
         return _cupy, "cupy"
     if backend == "auto":
         if cupy_gpu_available() and (problem_size is None
                                      or problem_size >= gpu_min_size):
             return _cupy, "cupy"
         return np, "numpy"
-    raise ValueError(
-        f"double_factorization error - unknown backend '{backend}'; expected "
-        "'auto', 'cupy', or 'numpy'.")
+    raise ValueError(f"unknown backend '{backend}'; expected 'auto', 'cupy', "
+                     "or 'numpy'.")
 
 
 def to_device(array: ArrayLike, xp: ArrayModule) -> DeviceArray:
--- a/python/cudaq_algorithms/double_factorization/_backend.py
+++ b/python/cudaq_algorithms/double_factorization/_backend.py
@@ -1,0 +1,111 @@
+# ============================================================================ #
+# Copyright (c) 2026 NVIDIA Corporation & Affiliates.                          #
+# All rights reserved.                                                         #
+#                                                                              #
+# This source code and the accompanying materials are made available under     #
+# the terms of the Apache License 2.0 which accompanies this distribution.     #
+# ============================================================================ #
+"""Array-backend selection for double factorization.
+
+Prefers the NVIDIA math libraries (cuSOLVER / cuBLAS via CuPy) when a GPU is
+available, and falls back to NumPy otherwise. All public double-factorization
+APIs accept a ``backend`` argument of ``"auto"`` (default), ``"cupy"``, or
+``"numpy"``.
+"""
+from __future__ import annotations
+
+import numpy as np
+
+try:  # CuPy is optional; absence simply forces the NumPy path.
+    import cupy as _cupy  # type: ignore
+except Exception:  # pragma: no cover - environment without CuPy
+    _cupy = None
+
+# Empirical CPU/GPU crossovers (orbital count ``n``) for ``backend="auto"``.
+# Below these the GPU's per-kernel launch + host-sync latency makes CuPy slower
+# than NumPy; above them the batched cuSOLVER/cuBLAS linear algebra wins and the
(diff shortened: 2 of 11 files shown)


kvmto (commented):
Reviewed the double factorization subpackage. The core math checks out end to end. I independently verified the C-DF gradient against central finite differences (matches to 1e-10 with and without regularization), the lstsq regularization normal equations, the CG versus lstsq equivalence, and the lcu one-norm, and all of these are correct. The questions below are the points I think are worth resolving before merge, ordered by severity. I could not test the CuPy GPU path here (no GPU available), so the one GPU question is unverified and flagged as such.

python/cudaq_algorithms/double_factorization/_factorization.py
```diff
@@ -0,0 +1,699 @@
...
+            # sum_{k<l} |Z_kl| (strict upper triangle) + 1/4 sum_k |Z_kk|
+            lam += float(np.sum(np.abs(np.triu(core, k=1))))
+            lam += 0.25 * float(np.sum(np.abs(np.diag(core))))
+        return lam
+
+    if convention == "burg":
+        for core in factorization.leaf_cores:
+            eigenvalues, vectors = np.linalg.eigh(np.asarray(core,
+                                                             dtype=float))
+            root = (vectors * np.sqrt(eigenvalues.astype(complex))) @ \
```
kvmto: The burg convention builds W from the symmetric matrix square root V sqrt(D) V^H. But (1/4) sum_i (sum_k |W_ki|)^2 is not invariant under the gauge freedom of Z = W W^T. The same core gives 1.554 with the symmetric square root, 1.519 with the eigen-column factor V sqrt(D), and 1.575 under an arbitrary right-orthogonal gauge. Should this use the eigendecomposition Z = sum_i lambda_i v_i v_i^T, giving (1/4) sum_i lambda_i (sum_k |v_ki|)^2 (the standard von Burg and Lee form)? As written, does the returned lambda match the paper's qubitization one-norm or a different convention?

python/cudaq_algorithms/double_factorization/_factorization.py
```diff
@@ -0,0 +1,699 @@
...
+        column = matrix[:, pivot_index].astype(float).copy()
+        for previous in vectors:
+            column = column - previous * float(previous[pivot_index])
+        vector = column / xp.sqrt(xp.asarray(pivot))
+        vectors.append(vector)
+        pivots.append(pivot)
+        residual = residual - vector * vector
+        most_negative = min(most_negative, float(xp.min(residual)))
+        residual = xp.where(residual > 0.0, residual, 0.0)
+    if initial_max > 0.0 and most_negative < -1.0e-8 * initial_max:
```
kvmto: The 'not positive semidefinite' warning is gated on initial_max > 0.0. For a negative-semidefinite supermatrix I get 0 leaves, reconstruction error around 12.5, and no warning at all, which is the case that most needs one. Should the warning (or a hard error) also fire when initial_max <= 0?

python/cudaq_algorithms/double_factorization/_factorization.py
```diff
@@ -0,0 +1,699 @@
...
+                                                   second_factor_threshold, xp)
+            rotations.append(rotation)
+            cores.append(core)
+            weights.append(pivot)
+    elif first_factorization == "eigendecomposition":
+        eigenvalues, eigenvectors = _sorted_symmetric_eigendecomposition(
+            supermatrix, xp)
+        abs_eigenvalues = to_numpy(xp.abs(eigenvalues))
+        for index in range(n * n):
+            if abs_eigenvalues[index] <= threshold:
```
kvmto: The Cholesky path has a numerical null-space floor of max(threshold, initial_max * 1e-14), but the eigendecomposition path only compares against threshold. With threshold=0.0 a true-rank-4 ERI produces 25 leaves (n^2 = 25) instead of 4. Is emitting the full n^2 zero-eigenvalue leaves intended, or should the eigen path get the same relative floor?

python/cudaq_algorithms/double_factorization/_factorization.py
```diff
@@ -0,0 +1,699 @@
...
+def _validate_eri(eri: ArrayLike) -> int:
+    eri = np.asarray(eri)
+    if eri.ndim != 4 or len(set(eri.shape)) != 1:
+        raise ValueError(
+            "double_factorization error - eri must be a square rank-4 tensor "
+            "(n, n, n, n) in chemist notation (pq|rs).")
+    # The leaf symmetrization silently assumes the real-orbital index
+    # symmetries within each pair; a violation would otherwise surface only
+    # as an unexplained reconstruction error.
+    if not (np.allclose(eri, eri.transpose(1, 0, 2, 3))
```
kvmto, line 170: This checks (pq|rs) == (qp|rs) and (pq|rs) == (pq|sr) but not the pair-swap (pq|rs) == (rs|pq). X-DF hides that by symmetrizing the supermatrix, but the C-DF -4 gradient prefactor folds four terms assuming full 8-fold symmetry on the un-symmetrized eri_dev. A 4-fold-only input passes validation and silently optimizes the wrong objective. Should this also check pair-swap symmetry, or should C-DF symmetrize its input first?

python/cudaq_algorithms/__init__.py
```diff
@@ -36,6 +39,9 @@
     _NATIVE_IMPORT_ERROR = exc
     __version__ = "CUDA-Q Algorithms (compiled extension not built)"
 
+# Pure-Python classical preprocessing (no compiled-extension dependency).
+from . import double_factorization
+
 # Pure-Python quantum primitives (no compiled-extension dependency).
 from . import (block_encoding, common_kernels, pauli_lcu, qsvt, qubitization,
```
kvmto, line 48: The double_factorization subpackage is cudaq-free in isolation, but this line imports the cudaq-requiring siblings right after it, so 'import cudaq_algorithms.double_factorization' still pulls in cudaq. Is the 'no cudaq dependency at all' guarantee meant to be reachable through an actual import path, or only structural?

python/cudaq_algorithms/double_factorization/_backend.py
```diff
@@ -0,0 +1,126 @@
...
+
+def expm_skew_symmetric_batched(generators: ArrayLike,
+                                xp: ArrayModule) -> DeviceArray:
+    """Batched :func:`expm_skew_symmetric` over a stack of real antisymmetric
+    matrices ``generators`` (shape ``(num_leaves, n, n)``). One batched Hermitian
+    eigendecomposition instead of ``num_leaves`` separate ones -- on the CuPy
+    backend this is a single cuSOLVER call rather than a Python-driven loop."""
+    generators = xp.asarray(generators)
+    hermitian = 1j * generators
+    eigenvalues, eigenvectors = xp.linalg.eigh(
```
kvmto, line 113: I could not test the CuPy path here, so this is a question rather than a claim. Does this batched eigh over the (num_leaves, n, n) stack route through cuSOLVER syevjBatched on CuPy, which historically caps at n <= 32? If so, does C-DF on GPU fail for more than 32 orbitals, which is inside the range where backend='auto' selects the GPU?

tests/python/test_double_factorization.py
```diff
@@ -0,0 +1,393 @@
...
+    # not a GPU is present): below the crossover the GPU loses.
+    assert resolve_backend("auto", problem_size=4, gpu_min_size=1000)[1] \
+        == "numpy"
+    # A large problem resolves to CuPy when a GPU is present, else NumPy.
+    expected = "cupy" if df.cupy_gpu_available() else "numpy"
+    assert resolve_backend("auto", problem_size=10000, gpu_min_size=1)[1] \
+        == expected
+
+
+def test_one_norm_conventions():
```
kvmto, line 269: These one-norm tests assert lambda >= base and monotonic decrease under regularization. Both hold even if the factorization value is wrong. Could we pin at least one lcu and one burg value against an independent hand computation, so a convention error like the one on the burg path would be caught?

tests/python/test_double_factorization.py
```diff
@@ -0,0 +1,393 @@
...
+    factorization = df.explicit_double_factorization(eri, threshold=0.0)
+    np.testing.assert_allclose(df.reconstruct_eri(factorization),
+                               factorization.reconstruct_eri())
+
+
+def test_modified_one_body_integrals():
+    eri = _synthetic_eri(n=4, num_vectors=2, seed=5)
+    one_body = np.diag([0.1, -0.2, 0.3, -0.4])
+    kappa = df.modified_one_body_integrals(one_body, eri)
+    expected = one_body - 0.5 * np.einsum("prqr->pq", eri)
```
kvmto, line 300: expected is computed with the byte-identical einsum 'prqr->pq' that the code uses, so no contraction-pattern error could ever fail this test. Should the oracle be an independent loop or explicit formula?

python/cudaq_algorithms/double_factorization/_factorization.py
```diff
@@ -0,0 +1,699 @@
...
+
+
+def _second_factorization(leaf: DeviceArray, scale: ArrayLike,
+                          second_factor_threshold: float,
+                          xp: ArrayModule) -> tuple[np.ndarray, np.ndarray]:
+    """Eigendecompose a symmetric leaf ``L = U diag(gamma) U^T`` and return
+    ``(U, Z)`` with the symmetric core ``Z = scale * outer(gamma, gamma)``."""
+    leaf = 0.5 * (leaf + leaf.T)
+    gamma, rotation = xp.linalg.eigh(leaf)
+    if second_factor_threshold > 0.0:
```
kvmto, line 150: This importance-weighted second-factor truncation branch (second_factor_threshold > 0) has no test. It works when I exercise it, but should there be a test so a regression cannot slip in unnoticed?

python/cudaq_algorithms/double_factorization/_backend.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+def to_numpy(array) -> np.ndarray:
+    """Return a host NumPy copy of a NumPy or CuPy array."""
+    if _cupy is not None and isinstance(array, _cupy.ndarray):
+        return _cupy.asnumpy(array)
+    return np.asarray(array)
+
+
+def expm_skew_symmetric(generator: ArrayLike, xp: ArrayModule) -> DeviceArray:
```
kvmto: Is expm_skew_symmetric (the non-batched version) dead code? I do not see it called, imported, or exported anywhere, only expm_skew_symmetric_batched is actually used.

python/cudaq_algorithms/double_factorization/_factorization.py
```diff
@@ -0,0 +1,699 @@
...
+    for rotation in xdf.leaf_rotations:
+        rotation = np.asarray(rotation, dtype=float).copy()
+        # exp(antisymmetric) is always special orthogonal (det +1), but eigh
+        # leaf rotations can be reflections (det -1), which have no real
+        # antisymmetric logarithm. Flipping one column flips the determinant
+        # without changing u_k u_k^T (hence without changing the factorization),
+        # so force det +1 before taking the logarithm.
+        if np.linalg.det(rotation) < 0.0:
+            rotation[:, 0] *= -1.0
+        skew = scipy.linalg.logm(rotation).real
```
kvmto, line 490: logm(rotation).real sits on a branch cut for a pi-rotation (eigenvalue near -1) and discards the imaginary part, so the generator may not exponentiate back to the rotation. This is warm-start only, so the impact is bounded, but would a cheap round-trip check be worth it?

python/cudaq_algorithms/double_factorization/_factorization.py
wsttiger: You're right, and thanks for the concrete gauge probe — reproduced exactly (three valid factors of the same core, three different values). The formula now uses the gauge-fixed eigendecomposition form you suggest: `1/4 sum_i |lambda_i| (sum_k |v_ki|)^2` per core, which is the standard von Burg/Lee convention and also what the leaf physically costs when implemented in its eigenbasis (`(1/2) sum_kl Z_kl n_k n_l = (1/2) sum_i lambda_i (sum_k v_ki n_k)^2`). For rank-one cores it still reduces to `(1/4)(sum_k |gamma_k|)^2`, so the X-DF behavior is unchanged; the docstring now states the convention and the gauge issue explicitly, and a new test pins a hand-computed value for a general (indefinite) core. Fixed in 6214434.

python/cudaq_algorithms/double_factorization/_factorization.py
wsttiger: Agreed — the gate skipped exactly the case that most needs it. The tracked minimum now includes the initial diagonal, and the warning threshold scales against `max(initial_max, -most_negative)`, so a negative-semidefinite supermatrix (0 leaves) warns loudly. Pinned by `test_negative_definite_supermatrix_warns_with_no_leaves`. Fixed in 6214434.

python/cudaq_algorithms/double_factorization/_factorization.py
wsttiger: Not intended — the eigendecomposition path now applies the same relative floor as Cholesky (`max(threshold, |lambda|_max * 1e-14)`), so `threshold=0` stops at the numerical rank. A rank-3 tensor now yields 3 leaves on both paths (test added). Fixed in 6214434.

python/cudaq_algorithms/double_factorization/_factorization.py
wsttiger: Your reading of the C-DF gradient is exactly right — the -4 prefactor folds four residual terms assuming full 8-fold symmetry on the unsymmetrized tensor. Between the two options we chose validation over silent symmetrization: `_validate_eri` now also requires `(pq|rs) == (rs|pq)`, keeping the documented 8-fold contract explicit rather than quietly repairing inputs that violate it (with a code comment explaining the C-DF rationale, and a test). Fixed in 6214434.

python/cudaq_algorithms/__init__.py
wsttiger: Structural only, as you suspected — the subpackage itself imports neither cudaq nor the extension, but every import path to it goes through the parent package, which loads cudaq. The docstring now says exactly that instead of overstating it. Fixed in 6214434.

python/cudaq_algorithms/double_factorization/_backend.py
wsttiger: We can't verify on GPU here either, but the concern matches the cuSOLVER documentation (`syevj`/`heevjBatched` documented for dimensions up to 32), and C-DF's `backend="auto"` does select GPU inside that risk range. `expm_skew_symmetric_batched` now falls back to per-leaf (non-batched, still-cuSOLVER) eigendecompositions on the CuPy path when n > 32, with the restriction documented in the docstring. Flagged as defensively handled but still unverified on actual hardware. Fixed in 6214434.

tests/python/test_double_factorization.py
wsttiger: Done — `test_one_norms_match_hand_computation` pins one lcu and one burg value against explicit hand arithmetic (a core built from hand-chosen eigenpairs, expected values written out in the test), so a convention error like the burg one now fails a test instead of passing the bounds. In 6214434.

tests/python/test_double_factorization.py
wsttiger: Agreed — the test now computes the expected correction with an independent explicit loop instead of the implementation's own einsum pattern. In 6214434.

python/cudaq_algorithms/double_factorization/_factorization.py
wsttiger: Test added: a data-driven threshold (median importance-weighted gamma) verifies truncation zeroes core diagonal entries and increases the reconstruction error, and an absurd threshold zeroes the cores entirely. In 6214434.

python/cudaq_algorithms/double_factorization/_backend.py
wsttiger: It was dead as written — but it turned out to be exactly the per-leaf fallback needed for your point about the CuPy batch cap (thread above), so it is now called from `expm_skew_symmetric_batched` for CuPy stacks with n > 32 and its docstring documents that role. In 6214434.

python/cudaq_algorithms/double_factorization/_factorization.py
wsttiger: Worth it, agreed — `_initial_generators` now round-trips `expm(logm(U).real)` against U and falls back to a neutral zero generator when they disagree (branch-cut case), so the optimizer is never seeded with a wrong warm start. In 6214434.

python/cudaq_algorithms/double_factorization/_backend.py
```diff
@@ -0,0 +1,137 @@
...
+    eigendecomposition instead of ``num_leaves`` separate ones -- on the CuPy
+    backend this is a single cuSOLVER call rather than a Python-driven loop.
+
+    CuPy dispatches stacked ``eigh`` to cuSOLVER's batched Jacobi solver
+    (``syevj``/``heevjBatched``), which is documented for matrix dimensions
+    up to 32; above that this falls back to per-leaf (non-batched)
+    eigendecompositions, which remain on the GPU through the standard
+    cuSOLVER path."""
+    generators = xp.asarray(generators)
+    if xp.__name__ == "cupy" and generators.shape[-1] > 32:
```
kvmto: Follow-up on this fallback, non-blocking. I got GPU access (RTX A6000, CuPy 13.6.0, CUDA 12.9) and tested the original concern directly. Batched cupy.linalg.eigh over a stacked (num_leaves, n, n) array returns correct results at n=33, 48, and 64 (matching scipy.expm to about 1e-12), with no crash and no silent error. CuPy's stacked eigh does not dispatch to the size-limited syevjBatched on this version. Measuring the guard: at n=48 with 16 leaves the per-leaf fallback is about 67 ms versus 31 ms for the batched path, so it is roughly 2.2x slower on exactly the large-n GPU case it was meant to protect, and this path runs every L-BFGS step. Since backend=auto selects the GPU for C-DF at n>=18, n>32 systems hit it. Suggestion: drop the fallback, or gate it behind an actual runtime capability or CuPy-version probe so current CuPy keeps the fast batched path. Keep it only if you need to support an older CuPy that genuinely hits the 32 limit. I could not find a version here that does, but I did not test older CuPy.

kvmto (approved):
Thanks for the thorough turnaround. I re-checked all eleven fixes against the code and by running them, and they hold up. The burg one-norm now uses the gauge-fixed eigendecomposition form (I confirmed it changes the value on indefinite cores and reduces to the old result for rank-one X-DF cores), and the eigendecomposition rank floor, the pair-swap validation, the negative-definite warning, and the logm round-trip all check out. The six new tests are strong, and the hand-computed one-norm test would fail the old broken formula. The full suite passes, including on GPU: nine cupy-parametrized tests ran on an RTX A6000 here. LGTM. One non-blocking follow-up on the n>32 fallback is in the inline note.

python/cudaq_algorithms/double_factorization/_backend.py
wsttiger: Thanks for actually testing it on hardware — that settles what we could only handle defensively. You're right on both counts: modern CuPy's stacked eigh doesn't hit the batched-Jacobi limit, and the guard was a 2.2x pessimization on every L-BFGS step in exactly the range `backend="auto"` picks the GPU. Dropped in 3b9fa8e: the fallback branch is gone, the now-unused unbatched `expm_skew_symmetric` is removed with it (which also lands your original dead-code point the direct way), and the batched docstring records your empirical verification instead of the cap warning. We didn't add a version probe since neither of us could find a CuPy that needs one.

PR #19, opened 2026-07-16
Add Givens-rotation Slater-determinant state preparation

wsttiger:
**Branch:** `wsttiger:features/add_givens_rotation_state_prep_python` →
base: PR #17 branch (`features/convert_stateprep_python`); retarget to
`main` once #17 merges. Sibling of the Hartree-Fock/UCC branch.

---

## Title

Add Givens-rotation Slater-determinant state preparation

## Body

This adds Slater-determinant preparation to `cudaq_algorithms.stateprep`:
given an orbital-coefficient matrix, a host-side pass compiles the
occupied orbitals into a schedule of adjacent Givens rotations, and
composable kernels apply that schedule to prepare the determinant —
real and complex variants.

### Surface

* **Kernels** (composable, register-first): `slater_determinant`,
  `complex_slater_determinant` (full preparation from a schedule's
  flattened arrays), and the primitive `givens_rotation` /
  `phase_givens_rotation` rotations they are built from. Adjacent
  rotations are realized as `exp_pauli` on contiguous three-qubit
  slices.
* **Host side**: `make_givens_rotation_schedule(orbital_coefficients,
  num_electrons, ...)` builds and validates a frozen
  `GivensRotationSchedule` (with `num_spin_orbitals` and an
  `is_complex` flag); `validate_givens_rotation_schedule`;
  `estimate_givens_resources` returning a frozen
  `GivensResourceEstimate` (echoes `num_spin_orbitals`/`num_electrons`
  plus rotation/gate-count proxies).
* **Injection**: `slater_determinant_kernel(schedule)` returns a
  `(qubits: cudaq.qview)` kernel directly usable as a `state_prep`
  argument to `PauliLCU.encode_kernel`, `Walk`, `QSVT`, and
  `Trotter.kernel` — pinned by a live injection test against the
  State-taking twin.
* For interleaved-spin systems the coefficient-matrix rows follow the
  package's alpha-even/beta-odd ordering (documented), so determinants
  compose with `hartree_fock_occupation` and the UCCSD conventions.

### Correctness

27 tests, validated against two independent references: the dense
minor-expansion oracle for determinant amplitudes, and a
second-quantized reference built from dense Jordan-Wigner creation
operators applied to the vacuum. Edge cases covered include single
orbital, full filling, zero-rotation schedules across the kernel
boundary, and complex schedules with zero phases reducing to the real
kernel. During the promotion from the prototype an estimator bug was
found and fixed (complex schedules were silently costed as real; the
prototype's test pinned the wrong behavior).

Example: `examples/stateprep/givens_slater_determinant.py`

### Test plan

* Full suite on the branch: 173 passed (base 145/146 + new tests).
* Example runs clean.


python/cudaq_algorithms/stateprep/_givens.py
```diff
@@ -0,0 +1,457 @@
...
+    if hasattr(orbital_coefficients, "tolist"):
+        is_complex = getattr(getattr(orbital_coefficients, "dtype", None),
+                             "kind", None) == "c"
+        rows = orbital_coefficients.tolist()
+    else:
+        is_complex = False
+        rows = [list(row) for row in orbital_coefficients]
+    if not is_complex:
+        is_complex = any(
+            isinstance(value, complex) for row in rows for value in row)
```
ivanbasov: `_as_matrix` can misclassify a genuinely complex matrix as real when it's passed as a nested Python list of `np.complex64` scalars. The numpy-array path uses `dtype.kind == 'c'` (line 164), but the list path relies on `isinstance(value, complex)` — and `np.complex64` is *not* a subclass of Python `complex` (only `np.complex128` is), so `is_complex` stays `False`. Concrete trigger: `make_givens_rotation_schedule([[np.complex64(np.cos(0.4))], [np.complex64(np.exp(1j*0.7)*np.sin(0.4))]])` — with `num_electrons == 1` there's no orthogonality check, the norm check passes, and the schedule is built on the real branch (lines 259-262), which drops the imaginary part of `lower`: the prepared determinant is silently wrong in magnitude and phase, no error raised. For `num_electrons >= 2` the same misdispatch instead computes the un-conjugated overlap (line 202) and raises a spurious "columns must be orthogonal" on a valid unitary. (numpy-array inputs and `.tolist()` outputs are unaffected, which is why the dispatch tests don't catch it.) Suggest detecting numpy complex explicitly, e.g. `isinstance(value, (complex, np.complexfloating))` (add `import numpy as np`), plus a complex64-nested-list dispatch test.

tests/python/test_stateprep_givens.py
```diff
@@ -0,0 +1,593 @@
...
+    state = np.zeros(dimension, dtype=complex)
+    state[0] = 1.0
+    for k in range(num_electrons - 1, -1, -1):
+        mode = sum(matrix[p, k] * creation(p)
+                   for p in range(num_spin_orbitals))
+        state = mode @ state
+    return state
+
+
+def _assert_allclose_up_to_global_phase(actual, expected, atol=1.0e-10):
```
ivanbasov: These statevector checks aren't precision-aware, so this file passes on the default `qpp-cpu` (fp64) but fails under the supported `CUDAQ_DEFAULT_SIMULATOR=nvidia` (fp32) target. Two spots: (1) `_assert_allclose_up_to_global_phase` defaults to `atol=1e-10` (with `rtol=1e-7`), and `test_givens_schedule_two_orbital_statevector` (line 155) uses that default against an fp32 simulator state whose accuracy floor is ~1e-6; (2) the many reference comparisons pinned at `atol=1e-6` (lines 164/176/185/224/234, spot-check 262) sit below the ~5e-6 error band a multi-rotation fp32 circuit reaches. The sibling `test_stateprep_kernels.py` already handles this: `tol = 5e-5 if np.dtype(cudaq.complex()) == np.dtype(np.complex64) else 1e-12`. Suggest adopting that precision-gated tolerance for the reference comparisons; the exact same-simulator comparisons (factory-vs-direct, zero-phase-vs-real, basis states) can stay at 1e-12.

ivanbasov (commented):
The Givens math checks out — validated against two independent references (dense minor-expansion and a second-quantized JW build) that agree with each other and the circuit, and the reversed-elimination order, complex per-rotation phases, and real-path pivot-sign omission are all correct. Two inline notes: one real (narrow) complex-dtype dispatch bug that silently prepares the wrong state, and a test-portability gap on the fp32 simulator target.

wsttiger:
Thanks @ivanbasov — the complex64 dispatch bug is a real one, nice find. Both addressed in fcbd465: the `np.complexfloating` dispatch fix + a nested-complex64 regression test, and precision-gated reference tolerances so the file is fp32-portable. Point-by-point in the threads.

python/cudaq_algorithms/stateprep/_givens.py
wsttiger: Confirmed and fixed. `np.complex64` isn't a subclass of Python `complex` (only `np.complex128` is), so `isinstance(value, complex)` left `is_complex` False and the schedule built on the real branch. Reproduced both failure modes — silent phase-drop at `num_electrons == 1`, spurious "columns must be orthogonal" at `num_electrons >= 2`. fcbd465 detects numpy complex explicitly (`isinstance(value, (complex, np.complexfloating))`, with `import numpy as np` added) and adds `test_dispatch_complex64_python_list_routes_complex`, which builds the schedule directly from a nested-`complex64` list (not through the complex128 helper, which would convert to an array and dispatch correctly regardless) and checks both `is_complex` and the prepared state.

tests/python/test_stateprep_givens.py
wsttiger: Fixed in fcbd465 by adopting the sibling file's precision-gated tolerance: `_REFERENCE_ATOL = 5e-5 if cudaq.complex() == complex64 else 1e-6`, applied to every simulator-vs-dense-reference comparison (including the two-orbital case that had used the `1e-10` default). The exact same-simulator comparisons (zero-phase-vs-real, basis states) and the pure-NumPy `second_quantized`-vs-`minors` check keep their tight `1e-12`. On fp64 (`qpp-cpu`) behavior is unchanged; I couldn't run the `nvidia` (fp32) target here without a GPU, but the gate mirrors `test_stateprep_kernels.py` exactly.

python/cudaq_algorithms/stateprep/_givens.py
ivanbasov: Confirmed — and verified beyond the test: I probed every input class against the fixed `_as_matrix` (nested complex64 lists, lists of 1-D complex64 arrays, mixed float32/complex64 rows, object-dtype arrays — all route complex; float32 inputs stay real, with thetas bit-identical to the fp64 build), reverted the fix to confirm the regression test fails first at `assert schedule.is_complex`, and grepped the package for other `isinstance(..., complex)`-style dispatch sites (none). Two comment-level nits, take or leave: the build's `tolerance=1e-5` is inert for this matrix — the float32 norm² of the [0.6, 0.8] column evaluates to exactly 1.0, so the schedule builds at the default tolerance and the "absorbs complex64 rounding" justification doesn't apply; and the docstring's `num_electrons >= 2` spurious-orthogonality mode can't trigger on these disjoint-support columns (the un-conjugated overlap is exactly zero). The dispatch assert pins the root cause regardless. Resolving.

tests/python/test_stateprep_givens.py
ivanbasov: Confirmed complete — I audited every comparison at HEAD: all 16 simulator-vs-reference sites use `_REFERENCE_ATOL`, and each comparison kept at 1e-12 is structurally exact even on fp32 (the complex-vs-real kernel pair differs only by `rz(0.0)`; the basis-state and factory-vs-direct checks run bit-identical circuits). I could also run what you couldn't: on the real `nvidia` (fp32) target the full file passes — 28 passed, worst simulator-vs-reference error 1.4e-7 — so 5e-5 has ~350× margin and the gate value is empirically right.

Two non-blocking follow-ups:

- Compute the gate at call time, inside `_assert_allclose_up_to_global_phase`, the way the sibling does. The module-scope constant is evaluated at import, and on a GPU machine with no `CUDAQ_DEFAULT_SIMULATOR` set the wheel's default target is `nvidia`, so `cudaq.complex()` is complex64 at import while conftest sets `qpp-cpu` (fp64) at runtime — fp64 runs silently get the 5e-5 gate. Verified on a GPU box: import-time dtype is complex64, complex128 only after `set_target("qpp-cpu")`.
- With that fixed, the fp64 branch can match the sibling's 1e-12 instead of 1e-6: measured on qpp-cpu, the worst error across this file's natively-built circuits is 2.7e-15 (~500× margin at 1e-12), while at 1e-6 an fp64 run can't catch the very bug class this PR fixes — a silent float32 truncation in host schedule math shows up at ~1e-8. The one exception is the new complex64-dispatch test: its reference is built from the float32-rounded (no longer exactly orthonormal) matrix, so it sits at a structural 1.9e-8 on fp64 — give it its own ~1e-7 atol and the file-wide fp64 branch tightens cleanly.

ivanbasov (approved):
Both addressed and verified. The dispatch fix is complete — I probed every input class against the fixed `_as_matrix`, confirmed the regression test fails on the un-fixed code, and found no other complex-dispatch sites in the package. The tolerance gate covers all 16 simulator-vs-reference sites, and I could run what you couldn't: the full file passes on the real `nvidia` fp32 target (28 passed; worst simulator-vs-reference error 1.4e-7, ~350× margin under 5e-5), so the gate value is empirically right. Two non-blocking follow-ups in the tolerance thread: compute the gate at call time (the import-time constant picks the fp32 branch on any GPU machine whose default target is `nvidia` while conftest runs qpp-cpu), and the fp64 branch can then be 1e-12 like the sibling (measured worst error 2.7e-15), with the new dispatch test getting its own ~1e-7 atol.

tests/python/test_stateprep_givens.py
wsttiger: Both taken in 7cbf5d5 — and thanks for running the fp32 target, good to have that empirically pinned.

- **Call-time gate**: moved the tolerance into a `_reference_atol()` helper computed inside `_assert_allclose_up_to_global_phase`, so it reflects the runtime target, not the import-time one. Confirmed the failure you described (import-time complex64 on a GPU box) is gone — on qpp-cpu `_reference_atol()` now returns 1e-12 at call time.
- **Tightened fp64 to 1e-12** (from 1e-6); all 28 pass, and it now catches the ~1e-8 float32-truncation class this PR is about. The one exception is exactly as you noted: the complex64-dispatch test's reference is built from the float32-rounded matrix (~1.9e-8 on fp64), so it keeps its own `atol=max(_reference_atol(), 1e-7)`.

python/cudaq_algorithms/stateprep/_givens.py
wsttiger: Thanks for the thorough verification (the input-class sweep and the reverted-fix check especially). Both nits fixed in 7cbf5d5: dropped the inert `tolerance=1e-5` from the dispatch test (you're right, the float32 norm of [0.6, 0.8] is exactly 1.0 so it builds at the default), and corrected the comment — it now just says the `is_complex` assert pins the root cause, without the `num_electrons >= 2` spurious-orthogonality claim (which, as you note, can't fire on these disjoint-support columns).

PR #20, opened 2026-07-16
Add Hartree-Fock and fixed-parameter UCC state preparation

wsttiger:
**Branch:** `wsttiger:features/add_hf_fixed_param_ucc_state_prep_python`
→ base: PR #17 branch (`features/convert_stateprep_python`); retarget to
`main` once #17 merges. Sibling of the Givens branch.

---

## Title

Add Hartree-Fock and fixed-parameter UCC state preparation

## Body

This adds Hartree-Fock reference preparation and fixed-parameter UCC
state preparation to `cudaq_algorithms.stateprep`, plus the package's
missing docs page.

### Surface

* **Kernels** (composable, register-first): `hartree_fock` (closed
  shell: fill the first `num_electrons` spin orbitals),
  `hartree_fock_occupation` (explicit occupied indices), and
  `fixed_parameter_ucc` (grouped Pauli words and coefficients at fixed,
  non-variational amplitudes over an HF reference — same shape as the
  `uccgsd`-family kernels, kept as its own named kernel for API clarity
  with arbitrary pools).
* **Host side**: `make_hartree_fock_occupation` for closed AND open
  shell (interleaved alpha-even/beta-odd layout, matching
  `get_uccsd_excitations` so determinants line up with pools built at
  the same `spin`); `validate_hartree_fock_occupation`;
  `get_fixed_parameter_ucc_pauli_lists(pool, num_qubits, tol)` with
  tolerance filtering and complex-coefficient rejection; resource
  estimators returning frozen `HartreeFockResourceEstimate` /
  `FixedParameterUccResourceEstimate` dataclasses.
* **Injection**: `hartree_fock_ucc_kernel(...)` returns a
  `(qubits: cudaq.qview)` kernel — HF reference plus the UCC product —
  directly usable as a `state_prep` argument to `PauliLCU`, `Walk`,
  `QSVT`, and `Trotter` factories (live-verified). The factory selects
  among four kernel shapes with positive guards so no empty list is
  ever closure-captured (cuda-quantum#4847).

### Correctness

15 tests: all prototype coverage ported, dense matrix-exponential
validation of the UCC kernel against the pool extended to open shell
((4,2,0), (6,3,1), (8,4,2)), and open-shell occupations cross-checked
against `get_uccsd_excitations`. Example:
`examples/stateprep/hartree_fock_ucc.py` (dense error ~3e-15).

### Test plan

* Full suite on the branch: 161 passed.
* Example and the docs page's injection snippet both run end-to-end.


python/cudaq_algorithms/stateprep/_hartree_fock.py
```diff
@@ -0,0 +1,308 @@
...
+            "provide exactly one of num_electrons or occupied_orbitals")
+    if occupied_orbitals is None:
+        occupation = make_hartree_fock_occupation(num_qubits, num_electrons,
+                                                  spin)
+    else:
+        if spin != 0:
+            raise ValueError(
+                "spin only applies with num_electrons; encode the open-shell "
+                "reference directly in occupied_orbitals")
+        occupation = [int(index) for index in occupied_orbitals]
```
ivanbasov: `hartree_fock_ucc_kernel` coerces `occupied_orbitals` with `int(index)` (line 266) *before* calling `validate_hartree_fock_occupation` (line 267), which defeats the "no silent coercion" contract `_as_count` enforces on the other entry points. `hartree_fock_ucc_kernel(4, [], [], [], occupied_orbitals=[0, 2.5])` truncates 2.5 -> 2 and silently prepares determinant {0, 2} with no error, whereas `estimate_hartree_fock_occupation_resources(4, [0, 2.5])` raises "non-negative integer"; `int(True)` slips a bool through the same way. Suggest validating the raw indices first, e.g. `occupation = [_as_count(i, "occupied orbital index") for i in occupied_orbitals]`, so a non-integral index is rejected rather than truncated into a wrong reference determinant.

python/cudaq_algorithms/stateprep/_hartree_fock.py
```diff
@@ -0,0 +1,308 @@
...
+    fixed-parameter UCCSD pool built at the same spin (e.g. 4 electrons at
+    spin 2 occupy ``{0, 1, 2, 4}``, not ``{0, 1, 2, 3}``).
+    """
+    num_qubits = _as_count(num_qubits, "num_qubits")
+    num_electrons = _as_count(num_electrons, "num_electrons")
+    spin_number = _as_count(spin, "spin")
+    if num_electrons > num_qubits:
+        raise ValueError("num_electrons cannot exceed num_qubits")
+
+    if spin_number == 0:
```
ivanbasov, line 102: `make_hartree_fock_occupation` accepts `(num_electrons, spin)` parity mismatches that its documented UCCSD sibling rejects, in two spots:
(1) At `spin=0` it short-circuits here (lines 102-103) and returns `list(range(num_electrons))` even for *odd* `num_electrons` — e.g. `make_hartree_fock_occupation(6, 3, 0)` -> `[0, 1, 2]` — while `get_uccsd_excitations(6, 3, 0)` raises the documented `RuntimeError("...number of electrons is odd but spin is 0")`. `[0, 1, 2]` is really an effective spin-1 (2 alpha + 1 beta) occupation, contradicting the docstring's "lines up with a UCCSD pool built at the same spin".
(2) For `spin > 0`, when `(num_electrons - spin)` is odd, `n_occ_beta = (num_electrons - spin) // 2` floors (line 111), so the realized `n_alpha - n_beta` is `spin + 1`, not `spin` — `make_hartree_fock_occupation(8, 4, 1)` returns `[0, 1, 2, 4]`, byte-for-byte identical to `spin=2`, with no error, even though the docs define `spin` as twice S_z (so it must share parity with the electron count).
Both let an HF reference build for a spin the matching pool would reject. Suggest one parity guard rejecting `(num_electrons - spin) % 2 != 0` (this also covers the odd-electrons-at-spin-0 case), matching `get_uccsd_excitations`. Low severity — only reachable with input that already violates the documented `spin` contract.

python/cudaq_algorithms/stateprep/_hartree_fock.py
```diff
@@ -0,0 +1,308 @@
...
+            or len(pauli_words) != len(coefficients)):
+        raise ValueError("parameters, Pauli-word groups, and coefficient "
+                         "groups must have the same length")
+    for words, coeffs in zip(pauli_words, coefficients):
+        if len(words) != len(coeffs):
+            raise ValueError("each Pauli-word group must match its "
+                             "coefficient group")
+        for word in words:
+            if isinstance(word, str):
+                if len(word) > num_qubits:
```
ivanbasov: `validate_fixed_parameter_ucc` only rejects *over-width* string words: the guard here is `len(word) > num_qubits`, but for a `num_qubits` register every valid word is exactly `num_qubits` wide (the `get_*_pauli_lists` helpers pad to full width via `term.get_pauli_word(num_qubits)`, and the kernel applies each word to the whole register). A short word slips through — `validate_fixed_parameter_ucc(4, [0.1], [["XY"]], [[1.0]])` passes, then fails later inside `exp_pauli` at kernel launch. Suggest `if len(word) != num_qubits:` so the host validator catches it with a clear message, matching the docstring's promise. Low severity — only hand-built words hit this, but that's exactly what this validator exists to catch.

docs/stateprep.md
```diff
@@ -0,0 +1,283 @@
...
+
+## Testing
+
+`tests/python/test_stateprep.py`, `test_stateprep_kernels.py`, and
+`test_operator_pools.py` pin the ansatz kernels and pools against the
+former compiled implementation (bit-identical circuits and operator
+coefficients). `test_stateprep_hf_ucc.py` validates the Hartree-Fock and
+fixed-parameter UCC surface against dense matrix exponentials of the
+operator pools, including open-shell occupations, input guards, and
+injection through the other primitives; `test_stateprep_givens.py` (on
```
ivanbasov: This says `test_stateprep_hf_ucc.py` validates "...and injection through the other primitives", but that file doesn't drive `PauliLCU.encode_kernel` / `Walk` / `QSVT` / `Trotter` with a `state_prep` — its factory tests only compare the factory kernel's state against the manual grouped-argument kernel. The end-to-end injection contract is actually exercised by `tests/python/test_state_prep_injection.py` (cited above at line 249) and, for Slater determinants, by `test_stateprep_givens.py::test_factory_injects_as_state_prep_kernel`. Suggest dropping "and injection through the other primitives" from the `test_stateprep_hf_ucc.py` clause, or adding an actual injection test to that file to match the claim.

ivanbasov (approved):
Approving — this is clean, well-structured, and the physics is correct and well-tested: the open-shell occupation math is cross-checked against `get_uccsd_excitations`, and `fixed_parameter_ucc` is validated against independent dense scipy `expm` of the pool at three (n, ne, spin) points. Nice touch making the kernel tests precision-aware. The four inline notes are optional, low-severity polish (a silent `int()` coercion, two `(num_electrons, spin)` parity-consistency gaps vs the UCCSD builders, an under-tight Pauli-word width check, and a docs overclaim about injection coverage) — none block merge.

wsttiger:
Thanks @ivanbasov! Took all four in 4aa7d1a — the `_as_count` validation for occupied indices, the open-shell parity guard, the exact-width Pauli check, and the docs fix. One nuance on the parity one below (scoped it to spin > 0 to keep the documented spin=0 behavior). Regression tests added.

python/cudaq_algorithms/stateprep/_hartree_fock.py
wsttiger: Fixed. `hartree_fock_ucc_kernel` now validates the raw indices with `_as_count(index, "occupied orbital index")` (which rejects non-integral values and bools) before building the occupation, instead of `int()`-truncating first. `occupied_orbitals=[0, 2.5]` and `[0, True]` now raise "must be a non-negative integer" rather than silently preparing a wrong determinant. Regression test added.

python/cudaq_algorithms/stateprep/_hartree_fock.py
wsttiger: Fixed the real half, scoped deliberately. Part (2) — the `spin > 0` floor bug where `(8, 4, 1)` silently aliased `(8, 4, 2)` — is now rejected: `make_hartree_fock_occupation` raises when `(num_electrons - spin)` is odd for `spin > 0` (placed after the `spin > num_electrons` check so that error still fires first). Regression test added.

On part (1) (odd electrons at `spin == 0`): I left that accepted, on purpose. The docstring documents `spin == 0` as "the contiguous set {0, ..., num_electrons - 1}" (the UCCSD-alignment promise is only in the `spin > 0` sentence), and it matches the `hartree_fock` device kernel, which fills the lowest `num_electrons` orbitals for odd counts too — an existing 1-electron test relies on this, and rejecting it would make the two inconsistent. Happy to tighten the docstring if you'd rather it spell out that `spin == 0` means "fill the lowest N", but I don't think it's an inconsistency as written.

python/cudaq_algorithms/stateprep/_hartree_fock.py
wsttiger: Fixed. `validate_fixed_parameter_ucc` now rejects `len(word) != num_qubits` (was `> num_qubits`), so a short word is caught on the host with a clear message instead of failing later inside `exp_pauli`. Regression tests for both over- and under-width words added.

docs/stateprep.md
wsttiger: Fixed — dropped the "injection through the other primitives" clause from the `test_stateprep_hf_ucc.py` sentence and pointed to `test_state_prep_injection.py`, which actually drives the `state_prep` contract.

python/cudaq_algorithms/stateprep/_hartree_fock.py
ivanbasov: Confirmed at HEAD — `[0, 2.5]` and `[0, True]` both raise "non-negative integer", and the fix is what the new test pins (with `int()` coercion, `[0, 2.5]` becomes the valid `[0, 2]` and the `pytest.raises` fails). One nit: the reply says both are regression-tested, but only `[0, 2.5]` is in the file — a one-line `[0, True]` case would pin the bool path against a future `_as_count` change. Non-blocking.

python/cudaq_algorithms/stateprep/_hartree_fock.py
ivanbasov: Agreed with the scoping — and it's safer than argued: for odd counts the contiguous fill coincides exactly with the spin=1 interleaved occupation (⌈n/2⌉ alphas on evens + ⌊n/2⌋ betas on odds; `hf(6,3,0) == hf(6,3,1) == [0,1,2]`, same at (10,5,·)), so it prepares the physically correct doublet, and every (num_electrons, spin)-parameterized pool builder raises at (6,3,0) — no silently-wrong-physics path exists. I'd take you up on the docstring sentence, e.g.: "`spin == 0` also accepts an odd `num_electrons`, meaning simply 'fill the lowest `num_electrons` spin orbitals' (matching the `hartree_fock` device kernel, and coinciding with the `spin == 1` open-shell occupation); pair such a reference with a pool built at the physical spin, since `get_uccsd_excitations` rejects odd counts at `spin == 0`."

One correction on the new guard's comment, though: "Matches get_uccsd_excitations" isn't quite right — the pool side has the identical un-guarded floor, so `get_uccsd_excitations(8, 4, 1)` is still accepted and returns exactly the (8, 4, 2) excitations (verified at HEAD; my round-1 "matching get_uccsd_excitations" was accurate only for the spin=0 rejection, so the phrasing is on me). `hartree_fock_ucc_kernel(num_electrons=4, spin=1)` is safe because your new HF guard fires first, but direct pool use (`make_uccsd_operator_pool(8, 4, 1)`) keeps the silent spin-2 aliasing. At minimum the comment should be corrected; the natural follow-up is mirroring the parity guard in `get_uccsd_excitations` — the same error-instead-of-silent-alias trade you made here, though it is a deliberate divergence from the C++ port's behavior, so your call whether that lands in this PR or a follow-up. Non-blocking.

docs/stateprep.md
ivanbasov: Verified both directions: `test_state_prep_injection.py` does drive the `state_prep` contract through PauliLCU/Walk/QSVT, and `test_stateprep_hf_ucc.py` doesn't touch primitive injection — dropping the clause was right. One nit on the new parenthetical: Trotter's injection test lives in `test_trotter.py` (`test_kernel_factory_accepts_state_prep_injection`), so "the primitives … covered by `test_state_prep_injection.py`" is slightly too broad; worth naming both files if you touch that sentence again. Non-blocking.

ivanbasov (approved):
Verified all four on this branch (15 tests pass; live probes: `[0, 2.5]` and `[0, True]` both rejected, (8, 4, 1) rejected while (8, 2, 3) keeps its original error, both word widths rejected, the docs clause checked against the actual test files). Your spin=0 scoping is right, and provably safe end-to-end: the odd-count contiguous fill coincides exactly with the spin=1 interleaved occupation, and every same-args pool builder raises — details in the thread. Still approved. Three non-blocking notes in the threads: the new guard's "Matches get_uccsd_excitations" comment isn't accurate (the pool side keeps the same un-guarded floor, so (8, 4, 1) still silently builds the spin-2 pool there — worth a mirrored guard as a follow-up), the `[0, True]` rejection isn't pinned by a test, and Trotter's injection coverage lives in `test_trotter.py`.

PR #21, opened 2026-07-16
Bridge double-factorized integrals to the LCU/qubitization/QSVT primitives

wsttiger:
## Summary

This PR closes the loop between the classical double-factorization
preprocessing and the quantum primitives: a new
**`cudaq_algorithms.chemistry`** module turns chemist-notation spatial
integrals into a qubit Hamiltonian (`cudaq.SpinOperator`) that feeds
`PauliLCU`, `Walk`, and `QSVT` directly.

1 commit on top of the double-factorization branch, ~330 insertions
across 5 files.

## Surface

```python
from cudaq_algorithms import PauliLCU, chemistry
from cudaq_algorithms import double_factorization as df

factorization = df.compressed_double_factorization(eri, num_leaves=T)
h = chemistry.qubit_hamiltonian(one_body,
                                df.reconstruct_eri(factorization),
                                scalar_offset=nuclear_repulsion)
encoding = PauliLCU(h)      # -> Walk / QSVT
```

- `chemistry.qubit_hamiltonian(one_body, eri, *, scalar_offset, tolerance)`
  — spin-expands the `(n, n)` core Hamiltonian and `(n, n, n, n)`
  chemist `(pq|rs)` tensor (interleaved spins) and applies the compiled
  Jordan-Wigner transform.
- `chemistry.spin_orbital_tensors(one_body, eri)` — the spin expansion
  by itself, for callers that need the fermionic tensors.

The module needs `fermion.jordan_wigner` only at call time; the
classical factorization itself has no such dependency. On current
`main` that transform lives in the compiled extension; once #15 (the
pure-Python fermion transforms) merges, the bridge picks up the pure
implementation with no code change here and the extension requirement
disappears.

## Why this matters

Compressing the factorization now shows up directly on the quantum
side: fewer leaves mean a smaller LCU normalization `alpha` (QSVT
degree for time evolution scales like `alpha * t`) and fewer Pauli
terms (SELECT cost), at the price of a spectrum shift controlled by the
tensor reconstruction error.

## Testing

5 tests (`test_df_qsvt_bridge.py`; they self-skip without the compiled
extension on current main, and run everywhere once #15 merges) validate
the whole chain on hardcoded H2/STO-3G integrals — no PySCF dependency:

- the bridged Hamiltonian reproduces the textbook FCI ground state to
  5e-5 Ha (the precision of the hardcoded integrals);
- a full-rank X-DF factorization round-trips the Hamiltonian (ground
  state and `alpha` to 1e-8);
- truncation lowers `alpha` and shifts the ground state by an amount
  bounded by the reconstruction error;
- the bridged Hamiltonian drives the qubitized walk (first moments
  against the dense matrix);
- shape validation.

## Example and docs

`examples/double_factorization/df_compression_to_qsvt.py` sweeps X-DF
leaf counts on H2/STO-3G and tabulates tensor error, `alpha`, term
count, and exact ground-state shift per truncation.


tests/python/test_df_qsvt_bridge.py
```diff
@@ -0,0 +1,104 @@
...
+
+import numpy as np
+import pytest
+
+import cudaq_algorithms as algorithms
+from cudaq_algorithms import PauliLCU, Walk, chemistry
+
+df = algorithms.double_factorization
+
+pytestmark = pytest.mark.skipif(
```
ivanbasov: The module-level skip guard is keyed to the wrong module. It gates on `find_spec("cudaq_algorithms._pycudaq_algorithms")`, but the bridge's actual runtime dependency is `fermion.jordan_wigner` (imported lazily in `qubit_hamiltonian`). These are different modules. On current main they coincide, so the guard looks correct today — but once the compiled extension is removed and `fermion` becomes pure Python, `find_spec("_pycudaq_algorithms")` is None on every install (the parent package still imports, so `find_spec` returns None rather than raising), making `pytestmark` a permanent `skipif(True)`. All five tests then get collected-then-skipped everywhere while CI stays green, dropping the H2 FCI check, the DF round-trip, the compression bound, and the Walk moment test exactly when the pure-Python bridge becomes exercisable everywhere — the opposite of the intended "runs everywhere" behavior. Note also that `test_bridge_validates_shapes` needs no extension at all (it raises before the deferred `from . import fermion`), yet it's gated too. Suggest keying the guard to the real dependency, e.g. `importlib.util.find_spec("cudaq_algorithms.fermion") is None`, or a try/except import of `fermion.jordan_wigner`, so it gates only environments where fermion is genuinely unavailable and stays correct across the fermion port.

python/cudaq_algorithms/chemistry.py
```diff
@@ -0,0 +1,100 @@
...
+    m = 2 * n
+    one_body_so = np.zeros((m, m), dtype=np.complex128)
+    two_body_so = np.zeros((m, m, m, m), dtype=np.complex128)
+    for p in range(n):
+        for q in range(n):
+            one_body_so[2 * p, 2 * q] = one_body[p, q]
+            one_body_so[2 * p + 1, 2 * q + 1] = one_body[p, q]
+            for r in range(n):
+                for s in range(n):
+                    coefficient = 0.5 * reordered[p, q, r, s]
```
ivanbasov, line 85: The spin expansion here is pinned to ground truth by a single scalar — the H2 minimum eigenvalue in `test_bridge_reproduces_h2_ground_state` (tol 5e-5) — and that assertion is insensitive to whole classes of index-transcription errors. Concretely, dropping the trailing `+1` on the beta-beta block (line 65: `two_body_so[2*p+1, 2*q+1, 2*r+1, 2*s]` instead of `...2*s+1`) yields a non-Hermitian operator whose minimum eigenvalue is unchanged to 6 digits, so this test still passes even though the full spectrum and the LCU 1-norm (PauliLCU `alpha`) are wrong. The other tests can't catch it either: the round-trip and compression tests apply the same expansion on both sides, and the Walk test compares `walk.moment` against `ket @ to_matrix() @ ket` for the same operator, so they're tautological w.r.t. an operator bug. Suggest at least one independent check: (a) assert Hermiticity of `to_matrix()` and compare a few excited eigenvalues or the LCU 1-norm to an independent reference, and (b) an n>=3 and/or a generic (non-H2) integral case so index/transpose typos that cancel under H2's symmetry are exposed. A direct-value assertion on `spin_orbital_tensors` output for a tiny hand-computed input (independent of the eigensolver), plus a complex-valued ERI case (the code coerces to complex128 unconditionally at lines 41-42 but only real input is exercised), would close most of the gap.

python/cudaq_algorithms/chemistry.py
```diff
@@ -0,0 +1,100 @@
...
+    spin orbitals (interleaved spins: ``2p`` up, ``2p + 1`` down), where
+    ``two_body_so[p, q, r, s]`` is the coefficient of
+    ``a^dag_p a^dag_q a_r a_s`` as consumed by ``fermion.jordan_wigner``.
+    """
+    one_body = np.asarray(one_body, dtype=np.complex128)
+    eri = np.asarray(eri, dtype=np.complex128)
+    n = one_body.shape[0]
+    if one_body.shape != (n, n):
+        raise ValueError("one_body must be a square (n, n) matrix")
+    if eri.shape != (n, n, n, n):
```
ivanbasov, line 58: `spin_orbital_tensors`/`qubit_hamiltonian` validate only tensor shapes (lines 44-49), not that `eri` obeys the 8-fold chemist-notation symmetry that Hermiticity of the output requires. A correctly-shaped but asymmetric `eri` (e.g. `eri[0,0,1,1]=1` with its symmetric partners unset) yields a non-Hermitian two-body coefficient tensor, so `jordan_wigner` returns a non-Hermitian SpinOperator that PauliLCU/Walk/QSVT silently consume as if Hermitian. The intended `reconstruct_eri` output is always symmetric so this never bites the documented pipeline, but since both functions are public it's worth either (a) stating the 8-fold-symmetry precondition explicitly in the docstring, or (b) adding a cheap, opt-out-able assertion (e.g. check `eri` against its `(1,0,3,2)`/`(2,3,0,1)`/`(3,2,1,0)` transposes) to turn a silent wrong result into a clear error. Note that a post-hoc Hermiticity check via `eigvalsh` would NOT catch this, since `eigvalsh` reads only one triangle.

python/cudaq_algorithms/chemistry.py
```diff
@@ -0,0 +1,100 @@
...
+# ============================================================================ #
+"""Chemistry-input bridges between classical tensors and qubit Hamiltonians.
+
+Connects the double-factorization preprocessing to the quantum primitives:
+chemist-notation spatial integrals (the form the DF module consumes and
+reconstructs) are spin-expanded and passed through the compiled
+``fermion.jordan_wigner`` transform, yielding a ``cudaq.SpinOperator``
+ready for ``PauliLCU``/``Walk``/``QSVT``.
+
+Requires the compiled extension (the ``fermion`` subpackage); importing
```
ivanbasov: Two inaccuracies in this docstring. (1) The import contract: `from . import fermion` is deferred inside `qubit_hamiltonian` (line 94), and the module itself imports only numpy, so `import cudaq_algorithms.chemistry` always succeeds — it never raises `ImportError`. "importing this module without it raises `ImportError`" (and the self-contradictory "at point of use" hedge) is therefore inaccurate; the error only surfaces when `qubit_hamiltonian` is actually called, and `spin_orbital_tensors` needs no fermion dependency at all, so "Requires the compiled extension" doesn't hold for the whole module. (2) Once the `fermion` subpackage is pure Python, calling it "the compiled extension" / "compiled Jordan-Wigner transform" is stale. The same wording recurs in double_factorization.md (lines 136 and 158 — "the compiled Jordan-Wigner transform", "requires the compiled extension") and example.py (lines 13, 19). Suggest fixing all three in one pass: describe the dependency as the `fermion` subpackage, note the `ImportError` is raised at the `qubit_hamiltonian` call rather than at import, and note `spin_orbital_tensors` on its own needs no fermion dependency.

python/cudaq_algorithms/chemistry.py
```diff
@@ -0,0 +1,100 @@
...
+    ``(n, n, n, n)`` chemist-notation ``(pq|rs)`` two-electron tensor over
+    real spatial orbitals — the exact convention the double-factorization
+    module documents. Returns ``(one_body_so, two_body_so)`` over ``2n``
+    spin orbitals (interleaved spins: ``2p`` up, ``2p + 1`` down), where
+    ``two_body_so[p, q, r, s]`` is the coefficient of
+    ``a^dag_p a^dag_q a_r a_s`` as consumed by ``fermion.jordan_wigner``.
+    """
+    one_body = np.asarray(one_body, dtype=np.complex128)
+    eri = np.asarray(eri, dtype=np.complex128)
+    n = one_body.shape[0]
```
ivanbasov: Scalar / 0-d `one_body` raises a confusing `IndexError: tuple index out of range` at `n = one_body.shape[0]` (line 43) instead of the intended `ValueError("one_body must be a square (n, n) matrix")`, because `n` is derived before the shape is validated. Guarding with a `one_body.ndim == 2` check (or reading the shape only after validating) would give a consistent error surface. Minor — other malformed inputs (1-D, size mismatch) are already handled.

ivanbasov (commented):
Nice, tight bridge — the chemist->second-quantized convention matches the tested spin-expansion recipe and the H2/STO-3G FCI check passes end-to-end. Two of the notes below are worth addressing before merge: the test file's skip guard is keyed to the compiled extension rather than the real `fermion` dependency (so it silently always-skips once the native build is retired), and the spin expansion is pinned only by the H2 minimum eigenvalue, which is blind to a class of index typos (e.g. a dropped `+1` that makes the operator non-Hermitian). The rest is low-severity polish.

wsttiger:
Thanks @ivanbasov — really sharp review, especially the skip-guard and the min-eigenvalue-blindness catches. All five in c122268. Verified the guard both ways (pure-Python fermion: 10 pass everywhere; compiled/absent: 3 pass, 7 skip) and confirmed the new Hermiticity check has teeth. Details in the threads.

tests/python/test_df_qsvt_bridge.py
wsttiger: Fixed, and thank you — this would have silently gutted the file post-#15. The guard now gates on the real dependency via a try/except import of `fermion.jordan_wigner` (rather than `find_spec`, since `fermion`'s package spec still resolves on this branch even though importing the compiled binding fails). Only the four operator-building tests carry `@_needs_fermion`; `test_bridge_validates_shapes` and the new `spin_orbital_tensors` tests are ungated (pure NumPy). Confirmed: pure-Python fermion -> all 10 run everywhere; compiled/absent -> those 3 run, the 7 fermion tests skip.

python/cudaq_algorithms/chemistry.py
wsttiger: Fixed with independent checks, using your dropped-`+1` example as the motivating case. Added: `test_bridged_operator_is_hermitian` (asserts `to_matrix()` is Hermitian — the typo yields a non-Hermitian operator whose min eigenvalue is unchanged to ~6 digits, so this catches that class directly; verified the assertion has teeth against a non-Hermitian matrix); `test_spin_orbital_tensors_direct_values` (a hand-computed assertion on the expansion output, independent of any eigensolver); `test_generic_symmetric_integrals_give_hermitian_operator` (a non-H2, n=3 case with generic Cholesky-symmetric integrals, so typos that cancel under H2's sparsity surface); and `test_complex_dtype_eri_matches_real_path` (exercises the unconditional complex128 coercion).

python/cudaq_algorithms/chemistry.py
wsttiger: Both, actually. `spin_orbital_tensors` now checks the three generators of the real-orbital chemist symmetry (`(pq|rs) = (qp|rs) = (pq|sr) = (rs|pq)`) and raises a clear error on an asymmetric `eri` — turning the silent non-Hermitian result into a failure — with an opt-out (`validate_symmetry=False`) for genuinely complex integrals whose symmetry differs. The precondition is now in the docstring, with a regression test using your `eri[0,0,1,1]=1` example. Confirmed `reconstruct_eri` output stays symmetric, so the documented pipeline is unaffected.

python/cudaq_algorithms/chemistry.py
wsttiger: Fixed across all three files. The module docstring now says `spin_orbital_tensors` is pure NumPy and always usable, and that `qubit_hamiltonian`'s `ImportError` (if any) surfaces at the call, not at import. Dropped "compiled extension" / "compiled Jordan-Wigner" here, in `docs/double_factorization.md` (both spots), and in the example.

python/cudaq_algorithms/chemistry.py
wsttiger: Fixed. `spin_orbital_tensors` now checks `one_body.ndim == 2` (and square) before deriving `n`, so a scalar/0-d input raises the intended `ValueError("one_body must be a square (n, n) matrix")`. Covered by the new rejection test.

tests/python/test_df_qsvt_bridge.py
ivanbasov: Confirmed both ways on this branch: compiled-absent → 3 pass + 7 skip; and with the pure-Python fermion from #15 grafted in (the post-merge future the old guard would have silently gutted) → all 10 run and pass. Two small notes: the reply's "only the four operator-building tests carry `@_needs_fermion`" is seven at HEAD — the three new Hermiticity/complex tests are gated too, correctly, since they call `qubit_hamiltonian`; just so readers don't expect them to run fermion-less. And the guard's `except Exception` means a present-but-broken fermion turns the 7 gated tests into silent skips rather than loud failures — `except ImportError` would gate absence while keeping unrelated breakage visible (my round-1 wording said just "try/except import", so the shape is on me). Non-blocking.

python/cudaq_algorithms/chemistry.py
ivanbasov: Confirmed — the dropped-+1 mutation that motivated this thread now fails 7 tests on this branch (was: passed everything), the direct-values assertions are all correct, and I re-derived the spin expansion against a first-principles dense-JW construction (agreement to 3e-16; the three symmetry generators do generate the full 8-fold group). Two residuals on the new tests' reach, both non-blocking:

- The reorder-typo class is still caught only by the fermion-gated energy tests. `test_spin_orbital_tensors_direct_values` probes only spatial (0,0,0,0), which every axis permutation fixes, and Hermiticity is structurally invariant under ANY reorder of an 8-fold-symmetric eri (the fermionic reversal conjugates into the symmetry group) — verified live: swapping the chemist reorder for the identity leaves 8 tests green and fails only the H2 FCI + compression tests, exactly the ones that skip on fermion-less installs. Cheap fix that also restores ungated coverage: assert a couple of mixed-spatial elements in the direct-values test, e.g. `two[0,0,2,2] == 0.5*ERI[0,1,0,1]` (exchange) and `two[0,2,2,0] == 0.5*ERI[0,0,1,1]` (Coulomb).
- `test_complex_dtype_eri_matches_real_path` compares bit-identical inputs: `astype(complex)` of real values converges with the real path at the complex128 coercion on the function's first lines (measured max difference: exactly 0.0), so genuinely complex-valued integrals stay unexercised. The meaningful variant works — I checked: a genuinely complex Hermitian eri (satisfying the complex chemist symmetry `(pq|rs)* = (qp|sr)`) is rejected by the real-orbital generator check — which incidentally proves the `validate_symmetry=False` opt-out is semantically necessary, as you argued — and with validation off yields a Hermitian Hamiltonian. Asserting that rejection + opt-out Hermiticity would cover what the dtype test currently can't.

ivanbasov (approved):
All five addressed and verified. The skip guard I checked both ways on this branch: compiled-absent gives 3 pass + 7 skip, and with the pure-Python fermion from #15 grafted in, all 10 run and pass — the post-merge scenario is safe now. The dropped-+1 mutation that motivated the review now fails 7 tests; I re-derived the spin expansion against an independent dense-JW construction (agreement to 3e-16), confirmed the three symmetry generators span the full 8-fold group, that `qubit_hamiltonian` forwards the validation, and that `reconstruct_eri` output sits ~6 orders below the validator's tolerance. Docs are clean of "compiled" in all three places. Approving; two non-blocking notes in the threads — the reorder-typo class is still visible only to the fermion-gated energy tests (a one-line strengthening of the direct-values test suggested), and the complex-dtype test compares bit-identical inputs (a verified stronger variant suggested).

PR #22, opened 2026-07-16
Add the double-factorized block encoding

wsttiger:
## Summary

This PR adds **`DoubleFactorizedEncoding`**: a second implementation of
the `BlockEncoding` protocol that encodes the electronic-structure
Hamiltonian directly from its double-factorized integrals (von Burg et
al., *PRX Quantum* **2**, 030305 (2021); arXiv:2007.14460), instead of a
flat Pauli expansion. `Walk` and `QSVT` consume it unchanged — this is
the payoff of the protocol-based dependency injection from #12.

1 commit on top of the double-factorization branch, ~1,300 insertions
across 5 files.

## Surface

```python
from cudaq_algorithms import DoubleFactorizedEncoding, Walk, QSVT
from cudaq_algorithms import double_factorization as df

factorization = df.compressed_double_factorization(eri, num_leaves=T)
encoding = DoubleFactorizedEncoding(one_body, factorization,
                                    scalar_offset=nuclear_repulsion)

walk = Walk(encoding)                    # same consumers as PauliLCU
kernel = QSVT(encoding).kernel(sequence)
```

## Construction

The factorized Hamiltonian is regrouped so every term is *diagonal in
some rotated orbital basis*:

```
H = const + sum_k F_k N_k + 1/2 sum_t sum_kl Z^t_kl (N^t_k - 1)(N^t_l - 1)
```

- **Frame 0**: the eigenbasis of the corrected one-body matrix (raw
  integrals + exchange correction + the remainder from centering the
  leaf number operators, all evaluated on the factorized tensor — a
  truncated factorization encodes exactly its truncated Hamiltonian).
- **One frame per leaf**, pure ZZ after centering.

SELECT applies ancilla-controlled Z words between **uncontrolled**
Givens-network frame changes (adjacent spatial rotations, lifted to
`XZY`/`YZX` `exp_pauli` pairs per spin on the interleaved Jordan-Wigner
register). With no control active the networks telescope to the
identity, so the controlled variants only control the Z words and sign
phases — the structural cost advantage of this encoding over a
Pauli-word SELECT. PREPARE and the walk composites reuse the `PauliLCU`
machinery unchanged.

**`alpha` is the published one-norm:** the centered construction makes
the encoding's subnormalization (minus the identity term) reproduce
`double_factorization_one_norm(..., "lcu")` exactly, pinned by a test.
Compressing the factorization lowers `alpha` and the term count
together — the knob a flat Pauli expansion does not have.

## Known limitation

`select_observable` raises `NotImplementedError`: the odd-Chebyshev-
moment observable is LCU-specific (it needs computational-frame Pauli
words). Even moments and every kernel factory work unchanged.

Both `encode_kernel` and `walk_kernel` support the package-wide
`state_prep` injection contract from #14: with `state_prep` (a
`(qubits: cudaq.qview)` kernel) they return zero-argument,
fully hardware-shaped circuits, in exact signature parity with
`PauliLCU` — pinned by twin tests against the `cudaq.State`-taking
forms plus sample smoke calls.

## Testing

**16 tests** (`test_df_encoding.py`), all against dense Jordan-Wigner
ladder references built in NumPy — no compiled extension, no chemistry
package:

- protocol conformance (`isinstance(..., BlockEncoding)`);
- encode block `== H / alpha` at 1e-12 (two seeds, 2 and 3 orbitals,
  scalar offsets, truncated factorizations);
- walk powers apply `T_k(-H/alpha)`; even Chebyshev moments through the
  `Walk` consumer; roundtrip identity;
- a mixed-direction QSVT sequence against the shared 2x2 signal model
  from the QSVT test suite;
- controlled-apply conventions (identity at control |0>, the encoding at
  control |1>);
- `alpha` against `double_factorization_one_norm`;
- validation and edge cases (identity-only encodings, diagonal systems,
  shape/symmetry rejection).

## Example and docs

`examples/double_factorization/df_block_encoding.py` compares the
encoding head-to-head with a `PauliLCU` of the same Hamiltonian (alpha,
terms, structure), sweeps factorization truncation, and measures
Chebyshev moments through the shared `Walk` consumer.


tests/python/test_df_encoding.py
```diff
@@ -0,0 +1,395 @@
...
+    np.testing.assert_allclose(block, (h @ ket) / encoding.alpha, atol=1e-12)
+
+
+def test_encode_block_three_orbitals():
+    one_body, eri = random_system(5, n=3)
+    encoding = DoubleFactorizedEncoding(one_body, eri)
+    h = dense_hamiltonian(one_body, eri)
+    ket = random_ket(64)
+    block = encoded_block(encoding, encoding.encode_kernel(), ket)
+    np.testing.assert_allclose(block, (h @ ket) / encoding.alpha, atol=1e-12)
```
eliotheinrich: The tests are nearly identical; fold together into a second pytest parameter on `n`?

python/cudaq_algorithms/df_encoding.py
```diff
@@ -0,0 +1,813 @@
...
+    "DoubleFactorizedEncoding",
+    "apply_basis_rotations",
+    "select",
+    "controlled_select",
+    "apply",
+    "walk",
+    "adjoint_walk",
+    "controlled_walk",
+    "controlled_adjoint_walk",
+]
```
eliotheinrich, line 78: Should there be an `__all__` here explicitly marking the internal kernel components as public if the only public element of the API is `DoubleFactorizedEncoding` (or at least that's my understanding)?

python/cudaq_algorithms/df_encoding.py
```diff
@@ -0,0 +1,813 @@
...
+                continue
+            for sigma in range(2):
+                for tau in range(2):
+                    a, b = 2 * k + sigma, 2 * l + tau
+                    pair = (a, b) if a < b else (b, a)
+                    words[pair] = words.get(pair, 0.0) + c
+    return constant, {
+        qubits: coeff
+        for qubits, coeff in words.items() if abs(coeff) >= threshold
+    }
```
eliotheinrich: threshold is enforced in two places; why not just
`if abs(c) > threshold: ...`
and then remove the extra gate at the end of the function?

tests/python/test_df_encoding.py
```diff
@@ -0,0 +1,395 @@
...
+_Z2 = np.diag([1.0, -1.0])
+_LOWER = np.array([[0.0, 1.0], [0.0, 0.0]])
+
+
+def _annihilator(mode: int, num_modes: int) -> np.ndarray:
+    ops = ([_Z2] * mode + [_LOWER] + [_I2] * (num_modes - mode - 1))[::-1]
+    out = np.array([[1.0]])
+    for op in ops:
+        out = np.kron(out, op)
+    return out
```
eliotheinrich, line 42: Duplicated from `df_encoding`; unless this is a deliberately different source of truth, probably makes sense to import them from there as needed

eliotheinrich (commented):
Look nice! Just a few small comments

python/cudaq_algorithms/df_encoding.py
```diff
@@ -0,0 +1,813 @@
...
+        # correction -1/2 sum_r (pr|rq) (which contracts the *factorized*
+        # tensor: U^T U = I collapses it to U diag(Z) U^T per leaf, so the
+        # encoded H is exactly the Hamiltonian of the reconstructed ERI,
+        # truncation included) plus the absorbed leaf singles
+        # sum_l Z^t_kl from centering the number operators (see
+        # ``_leaf_pauli_expansion``).
+        kappa = kappa_base.copy()
+        for rotation, core in zip(factorization.leaf_rotations,
+                                  factorization.leaf_cores):
+            absorbed = core.sum(axis=1) - 0.5 * np.diag(core)
```
ivanbasov: The kappa centering and the leaf ZZ expansion use inconsistent core symmetrizations. Line 445 absorbs `core.sum(axis=1)` (row sum only), but `_leaf_pauli_expansion` (line 358) builds its ZZ words from the symmetrized `0.125*(core[k,l]+core[l,k])`. Since the physical leaf `1/2 Σ_kl Z_kl N_k N_l` depends only on the symmetric part of the core, the correct one-body centering coefficient is `Σ_l Zsym[k,l] = ½(rowsum_k + colsum_k)`. For any symmetric core these coincide, so every factorization from explicit_/compressed_double_factorization and the 1e-12 encode==H tests are unaffected. But a non-symmetric leaf core handed in via a user-built DoubleFactorization would encode the wrong operator, off by ½(colsum−rowsum) in the one-body part, with no error surfaced (the constructor validates one_body symmetry at line 412 but not leaf_cores). Suggest either symmetrizing the absorbed vector, e.g. `absorbed = 0.5*(core.sum(axis=1) + core.sum(axis=0)) - 0.5*np.diag(core)`, or asserting/documenting that leaf cores must be symmetric.

python/cudaq_algorithms/df_encoding.py
```diff
@@ -0,0 +1,813 @@
...
+                frames.append((_givens_sweep(rotation), leaf_words))
+
+        self._constant = constant
+
+        # The identity total is one frameless term; it lives in the first
+        # frame's group (rotations act trivially on it).
+        terms: list[tuple[float, tuple[int, ...], int]] = []
+        if abs(constant) >= coefficient_threshold:
+            terms.append((constant, (), 0))
+        frames = [(sweep, words) for sweep, words in frames if words]
```
ivanbasov: `num_frames`/`terms` don't honor the documented 1:1 frame↔leaf correspondence when a frame's words are all elided. The filter at line 478 (plus the `if leaf_words:` guard at 468) drops any empty-word frame, and both `num_frames` and the `frame_index` stored in `terms` are re-derived from the surviving frames only. So the invariant `num_frames == 1 + factorization.num_leaves` (asserted in test_satisfies_block_encoding_protocol) and the docstring/df_block_encoding.md claim that frame 0 is the kappa eigenbasis both break when frame 0's one-body words are all filtered out. Concrete trigger via the public constructor: pick `one_body` that cancels the absorbed leaf-singles correction so `kappa == 0` (then `one_body_words` is empty and frame 0 is dropped), or set `coefficient_threshold` above every one-body coefficient. The first surviving leaf frame is then re-tagged `frame_index 0`, and leaf indices shift down by one. Not a correctness bug in H/alpha — the eigenvalue energy is already folded into `constant` (line 457) and the current tests only use `random_system`, where frame 0 is always retained. Suggest either keeping a zero-term placeholder frame for the kappa eigenbasis (preserving the count/index invariant) or softening the docstring, df_block_encoding.md, and the test to state that fully-thresholded frames are elided.

tests/python/test_df_encoding.py
```diff
@@ -0,0 +1,395 @@
...
+        if flip == 1:
+            x(control_and_ancilla[0])
+        controlled(control_and_ancilla, system)
+        if flip == 1:
+            x(control_and_ancilla[0])
+
+    return circuit
+
+
+def test_controlled_apply_control_conventions():
```
ivanbasov, line 268: `controlled_walk_step_kernel` and `controlled_adjoint_walk_step_kernel` (and the underlying `controlled_walk` / `controlled_adjoint_walk`, both in `__all__`) have no direct test. `test_controlled_apply_control_conventions` only routes through `controlled_apply_kernel` -> `controlled_select`, so the `controlled_select + controlled_reflect_about_prepare` composition — and its reversed order in the adjoint variant — is never asserted. The atomic pieces are covered (controlled_select at both control values here, controlled_reflect_about_prepare in pauli_lcu), so this is low-risk, but a small direct test would pin the glue cheaply: for both factories, assert the controlled walk step is identity on system+ancilla at control |0> and equals the uncontrolled walk / adjoint-walk step at control |1> (a controlled-roundtrip == identity check at control 0 and 1 exercises both in one shot).

python/cudaq_algorithms/df_encoding.py
```diff
@@ -0,0 +1,813 @@
...
+        rot_ptr += segment_rot_counts[f]
+        for _ in range(frame_term_counts[f]):
+            for b in range(n_anc):
+                if term_controls[ptr_ctrl] == 0:
+                    x(ancilla[b])
+                ptr_ctrl += 1
+            for _ in range(term_lengths[term_i]):
+                z.ctrl(ancilla, system[term_targets[ptr_tgt]])
+                ptr_tgt += 1
+            if term_signs[term_i] < 0:
```
ivanbasov, line 146: The single-ancilla negative-sign branch here (`if n_anc == 1: z(ancilla[0])`) has no test coverage. It fires only when a term has a negative coefficient and `num_ancilla == 1` (i.e. `len(terms) <= 2`, from `max(1, (len(terms)-1).bit_length())`). The one single-ancilla test, `test_identity_only_encoding`, uses `scalar_offset=+1.5`, so `term_signs == [+1]` and this branch is skipped; all other tests have `num_ancilla >= 2` and take the multi-control `z.ctrl` branch. Since this is a distinct special case (added to avoid the empty-control-list crash of `ancilla.front(0)`, cuda-quantum#4847), a sign error would go uncaught. One-line closer: add `DoubleFactorizedEncoding(np.zeros((2,2)), np.zeros((2,2,2,2)), scalar_offset=-1.5)` and assert `block == -ket` (ideally covering the `controlled_select` twin too).

python/cudaq_algorithms/df_encoding.py
```diff
@@ -0,0 +1,813 @@
...
+        if kappa_base.ndim != 2 or kappa_base.shape[0] != kappa_base.shape[1]:
+            raise ValueError("one_body must be a square (n, n) matrix")
+        if not np.allclose(kappa_base, kappa_base.T, atol=1e-10):
+            raise ValueError("one_body must be symmetric")
+        n = kappa_base.shape[0]
+
+        if isinstance(two_body, DoubleFactorization):
+            factorization = two_body
+        else:
+            eri = np.asarray(two_body, dtype=float)
```
ivanbasov, line 419: Input validation is one-sided: `one_body` is checked for squareness and symmetry (raising `ValueError`), but a raw `two_body` ERI is only shape-checked `(n,n,n,n)` before being passed to `explicit_double_factorization(..., threshold=0.0)` — its chemist 8-fold index symmetry is never validated. An ERI violating that symmetry (e.g. `eri[p,q,r,s] != eri[q,p,r,s]`) is accepted; if the factorizer symmetrizes via an eigendecomposition (reads one triangle), the encoding block-encodes a symmetrized operator that differs from the tensor the caller passed, silently. This path is also untested — `random_system` always builds a symmetric ERI. Consider either asserting the ERI index symmetry here (fast-fail, matching the one_body check) or confirming `explicit_double_factorization` already validates it. Low priority; malformed input only.

ivanbasov (approved):
Approving — this is the deepest and best-tested module in the set, and it reads as correct: the block encoding is pinned by an independent dense Jordan-Wigner reference (encode block == H/alpha to 1e-12 at 2 and 3 orbitals, with offset and a truncated factorization), plus alpha vs double_factorization_one_norm, walk-power vs Chebyshev, a QSVT-vs-signal-model sequence, and the controlled-apply / roundtrip / identity-only / diagonal edges. The five inline notes are all low-severity and non-blocking: a latent kappa-vs-leaf symmetrization inconsistency that only bites a hand-built non-symmetric leaf core, a num_frames/terms metadata invariant that breaks only when a frame is fully thresholded away, an unvalidated ERI index symmetry, and two cheap coverage gaps (the controlled walk-step kernels and the single-ancilla negative-sign SELECT branch).

wsttiger:
Thanks both — @ivanbasov and @eliotheinrich. All nine in ae01c2d: the leaf-centering symmetrization and the frame-invariant placeholder (both with tests that fail without the fix), the ERI symmetry guard, direct tests for the controlled walk step and the single-ancilla sign branch, plus the threshold consolidation and test fold-in. Two I answered rather than changed (the `__all__` question and the "duplicated reference" note) — reasoning in the threads.

python/cudaq_algorithms/df_encoding.py
wsttiger: Fixed. The absorbed one-body vector is now `0.5*(core.sum(axis=1) + core.sum(axis=0)) - 0.5*diag(core)`, matching the symmetrized core the ZZ expansion uses. No-op for symmetric cores (so every factorizer output and the 1e-12 tests are unchanged), correct for a user-built non-symmetric core. Added `test_nonsymmetric_leaf_core_matches_symmetrized` (encoding of a non-symmetric core == its symmetrized version) — confirmed it fails on the old row-sum-only form.

python/cudaq_algorithms/df_encoding.py
wsttiger: Fixed by keeping placeholders (your first option). Every frame — frame 0 and each leaf — is retained even if all its words threshold away; an empty frame contributes no SELECT terms and its rotation sweep telescopes to the identity. So `num_frames == 1 + num_leaves` and "frame 0 = kappa eigenbasis" stay literally true, and the docstring/md need no softening. Added `test_num_frames_invariant_when_kappa_frame_is_elided` (one_body chosen so `kappa == 0`, so frame 0's words all vanish) — confirmed it fails without the placeholder.

python/cudaq_algorithms/df_encoding.py
wsttiger: Fixed — same fast-fail as the `one_body` check right above: the constructor now rejects a raw ERI that violates the chemist permutation symmetry (`(pq|rs) = (qp|rs) = (pq|sr) = (rs|pq)`). (Consistent with the same fix on PR #21's `chemistry.spin_orbital_tensors`.)

tests/python/test_df_encoding.py
wsttiger: Added `test_controlled_walk_step_roundtrip_is_identity`, parametrized on the control value: `controlled_walk_step` then `controlled_adjoint_walk_step` is identity at control |1> (W then W-dagger) and at control |0> (both identity), pinning `controlled_walk` / `controlled_adjoint_walk` and their reversed segment order in one shot, exactly as you suggested.

python/cudaq_algorithms/df_encoding.py
wsttiger: Added `test_single_ancilla_negative_sign_select`: your one-liner (`scalar_offset=-1.5`, zero integrals) gives `num_ancilla == 1`, a negative term, and `block == -ket`, exercising the `n_anc == 1` `z(ancilla[0])` branch (and its `controlled_select` twin via the controlled tests).

python/cudaq_algorithms/df_encoding.py
wsttiger: Consolidated to a single inline `abs(c) >= threshold` check and dropped the end-of-function filter. Confirmed it's safe: each `(k, l)` contributes to distinct qubit pairs (diagonal -> `(2k, 2k+1)`; off-diagonal -> four `(2k+s, 2l+t)` with the second index > `2k+1`), so no word ever accumulates from more than one term — the inline and end-filter forms are equivalent.

tests/python/test_df_encoding.py
wsttiger: Folded `test_encode_block_three_orbitals` into `test_encode_block_is_h_over_alpha`, now parametrized on `(seed, n)` with the ket dimension derived from `n`.

python/cudaq_algorithms/df_encoding.py
wsttiger: Kept as-is, deliberately. The composable device kernels (`select`, `walk`, `controlled_walk`, ...) are meant to stay reachable in the module namespace so they can be composed inside user kernels — the same convention the rest of the package follows (e.g. `pauli_lcu`). The public boundary is drawn at the *package* level: only `DoubleFactorizedEncoding` is re-exported from `cudaq_algorithms`, while these stay one level down on purpose. So the existing `__all__` (which lists them) is intentional, not an oversight.

tests/python/test_df_encoding.py
wsttiger: This is the "deliberately different source of truth" case you flagged: the dense Jordan-Wigner ladder reference (`_annihilator` / `dense_hamiltonian`) is intentionally independent of the encoding under test — that independence is the point of the file (see its module docstring). `df_encoding` has no dense-ladder helpers to import from; it builds circuits, not dense matrices. Kept separate on purpose.

PR #23, opened 2026-07-28
Test wheels across arm64/x86 × CPU/GPU, and collapse to one CUDA-agnostic wheel

wsttiger:
## What

Two things, both validated end to end:

1. **Test the wheels across `{arm64, x86} × {CPU, GPU}`** — a 4-leg matrix, so the production wheels are exercised on every platform QA runs.
2. **Ship a single CUDA-agnostic wheel** — depend on the `cudaq` metapackage instead of `cuda-quantum-cuNN`, dropping the per-CUDA split and our own selector metapackage (thanks @bmhowe23).

Also brings the pure-Python wheels workflow to `main` and moves the CUDA-Q pin to the 0.15 line.

## Packaging: one wheel

- The package is pure Python, so it depends on **`cudaq >= 0.15.0, < 0.16`** (the CUDA-agnostic metapackage, which selects `cuda-quantum-cuNN` at install) and ships one `cudaq_algorithms-<v>-py3-none-any.whl`.
- **Deleted**: `cudaq-algorithms-cu12/-cu13`, both `pyproject.toml.cu*` variants, the `python/metapackages/` selector, and `build_metapackages.sh` (net −410 lines).

## Testing

- **`build_cudaq_wheels`** builds CUDA-Q from the pinned `.cudaq_version` commit for arm64 + x86 and cu12 + cu13 (Custom mode).
- **`test-wheels`** matrix `{amd64, arm64} × {cpu, gpu}`: CPU legs run `qpp-cpu` (cu12); GPU legs run in the CUDA-Q **devcontainer** with `--gpus all` + `nvidia-fp64` (cu13) — fp64 because the suite's 1e-10..1e-12 tolerances need it, and the devcontainer supplies the CUDA toolkit CuPy needs.
- **`cudaq_wheels` toggle** (default `Custom`):
  - **Custom** — pre-install the from-source **unreleased** `cuda-quantum-cuNN==0.15.99`, then `pip install --no-deps cudaq-algorithms` (bypasses the `cudaq` resolver; tests against the pinned commit).
  - **PyPI** — `pip install cudaq-algorithms` resolves `cudaq` → a released cuda-quantum (the real end-user path).

## Validation

Custom-mode dispatch `30422991006`: the single wheel built once, all four CUDA-Q builds, and all four test legs green — the wheel installs against the from-source `cuda-quantum-cuNN==0.15.99` and the suite passes on CPU (`qpp-cpu`) and GPU (`nvidia-fp64`, CuPy included).

## Notes

- GPU legs test **cu13** (the GPU runners are CUDA-13 hosts); cu12 is covered on CPU.
- Follow-up (out of scope): a macOS-14 hosted CPU smoke job.


.github/workflows/wheel_algorithms.yaml
```diff
@@ -112,42 +132,20 @@ jobs:
...
         run: |
           git config --global --add safe.directory "$GITHUB_WORKSPACE"
           bash scripts/ci/build_wheels.sh \
             --cuda-version ${{ steps.config.outputs.cuda_major }} \
-            --cudaq-prefix /usr/local/cudaq \
-            --build-type ${{ inputs.build_type }} \
             --python-version ${{ matrix.python }} \
-            --version ${{ inputs.version || '0.14.99' }} \
-            --devdeps
+            --version ${{ inputs.version || '0.15.99' }}
```
ivanbasov, line 124: Non-blocking: dropping `--build-type` here removed the last reference to `inputs.build_type` — it's still declared in both trigger blocks (a `required: true` choice on `workflow_dispatch`), and `pr_workflow.yaml` still passes `build_type: Release`, but nothing consumes it anymore. Someone dispatching a `Debug` run for triage will silently get wheels identical to the Release default. (It was already functionally inert — `build_wheels.sh` parses `--build-type` and never uses it for the pure-Python wheel — but now it's syntactically dead too.)

Suggestion: drop the input from both trigger blocks together with the `build_type: Release` line in `pr_workflow.yaml` (they need to land together — passing an undeclared input to a reusable workflow fails validation), or keep it and note in its description that the pure-Python wheels ignore it.

.github/workflows/wheel_algorithms.yaml
```diff
@@ -192,20 +190,70 @@ jobs:
...
       contents: read
     strategy:
       fail-fast: false
       matrix:
-        python: ['3.12']
-        platform: ['amd64']
-        cuda_version: ['12.6']
+        # Each leg is fully specified: the device only changes the runner and
+        # the simulator, never the artifacts (the cuda-quantum-cuNN wheel
+        # carries both the qpp-cpu and nvidia simulators).
```
ivanbasov: Non-blocking, comment accuracy: with the legs pinned as they are, `device` is coupled 1:1 to the CUDA major — both `cpu` legs download only the cu12 artifacts and both `gpu` legs only the cu13 ones (both download steps key on `cuda_major`), plus the container changes. So "the device only changes the runner and the simulator, never the artifacts" reads stronger than what the matrix actually does; same for "built once and consumed by every wheel-test leg" on the artifact-name comments — each per-CUDA-major wheel feeds 2 of the 4 legs. The sibling comment in `build_cudaq_wheels.yaml` ("CPU hosts fall back to cu12") has the same flavor: at test time nothing falls back — the CPU legs pin `cuda_version: '12.6'` explicitly.

The substantive flip side is a small coverage split: the cu13 wheel is never exercised with `qpp-cpu`, and the cu12 wheel never with `nvidia-fp64`, so the parenthetical ("the cuda-quantum-cuNN wheel carries both the qpp-cpu and nvidia simulators") isn't actually validated in both directions by this matrix. That seems like a perfectly reasonable trade-off for a smoke matrix — just worth stating the pairing as deliberate, e.g.: "device selects the runner, container, simulator and — as the legs are pinned — the CUDA major (cpu -> cu12, gpu -> cu13)".

.github/workflows/wheel_algorithms.yaml
```diff
@@ -93,11 +111,13 @@ jobs:
       matrix:
         python: ['3.12']
         platform: ['amd64']
```
ivanbasov, line 101: Non-blocking, latent trap: this job keeps the `platform` matrix axis (after this PR it feeds only `runs-on`) while the uploaded artifact name below became arch-neutral. The natural future edit — adding `'arm64'` here, mirroring what this PR does in `build_cudaq_wheels.yaml` — would make the paired legs upload to the same `algorithms-wheels-py...-cuNN` name, and `upload-artifact@v4` rejects duplicate names within a run (409), so the slower leg would fail CI even though the wheels are byte-identical `py3-none-any`.

Since the wheel is arch-independent and built once by design, consider deleting the axis entirely (hardcode `linux-amd64-cpu8` in `runs-on`, keeping the fork fallback) so the workflow structurally can't grow a colliding second leg.

pyproject.toml.cu12
```diff
@@ -12,7 +12,7 @@ requires-python = ">=3.11"
 readme = "README.md"
 license = "Apache-2.0"
 dependencies = [
-  "cuda-quantum-cu12 == 0.14.*",
+  "cuda-quantum-cu12 >= 0.15.1, < 0.16",
```
ivanbasov: Non-blocking (dispatch-only), but probably worth a decision before merge: as of today this bound makes PyPI mode unsatisfiable. `cuda-quantum-cu12/cu13 >= 0.15.1, < 0.16` matches nothing on either index pip consults — pypi.org tops out at 0.15.0 for both cu12 and cu13, and pypi.nvidia.com carries only 0.14.0. In PyPI mode there is no fallback: `build-cudaq-wheels` and the download step are skipped and `test_wheels.sh` gets `SKIP`, so the `cudaq-algorithms` install ends in `ResolutionImpossible` on all four legs. Custom mode masks this because the from-source wheels are stamped 0.15.99. (Related: the toggle's description says "released wheels from pypi.nvidia.com", but with no 0.15.x on that index, resolution would actually come from pypi.org once a matching release exists.)

If the `>= 0.15.1` floor is intentional (the pinned `.cudaq_version` commit is post-0.15.0, so released 0.15.0 may lack what you need), consider noting in the toggle's description that PyPI mode becomes usable once cuda-quantum 0.15.1 ships — and giving it one validation dispatch then. If released 0.15.0 is actually sufficient, relaxing the floor to `>= 0.15.0` would make the mode testable today.

.github/workflows/wheel_algorithms.yaml
```diff
@@ -192,20 +190,70 @@ jobs:
...
-    name: Test CUDA-Q Algorithms wheels (CPU)
+  test-wheels:
+    name: Test CUDA-Q Algorithms wheels (${{ matrix.platform }}, ${{ matrix.device }})
     needs: [build-cudaq-wheels, build-algorithms-wheels, build-metapackages]
-    if: ${{ !failure() && !cancelled() }}
-    runs-on: ${{ startsWith(github.repository, 'NVIDIA/cudaq-algorithms') && format('linux-{0}-cpu4', matrix.platform) || 'ubuntu-latest' }}
-    container: ubuntu:24.04
+    # Consumes the from-source CUDA-Q wheels, so it can only run where
+    # build-cudaq-wheels ran (skipped on forks along with it).
+    if: ${{ !failure() && !cancelled() && startsWith(github.repository, 'NVIDIA/cudaq-algorithms') }}
```
ivanbasov: Non-blocking, stale comment: this was true before the `cudaq_wheels` toggle, but `build-cudaq-wheels` is now also skipped in PyPI mode (its own comment got updated accordingly), while this one still promises "it can only run where build-cudaq-wheels ran". In PyPI mode this job runs precisely when that job was skipped, consuming released wheels instead — which is exactly what the `!failure() && !cancelled()` guard makes possible. Someone "tightening" the guard to `success()` on the strength of this comment would silently break PyPI mode; suggest describing both modes here.

ivanbasov (approved):
Reviewed the full PR at e331d44 — the two workflow files plus the scripts, action, and pyproject variants the last two commits pulled in — and cross-checked the plumbing end to end: all four legs' downloads resolve to artifacts that are actually built ({amd64, arm64} x {cu12, cu13} CUDA-Q wheels; per-CUDA-major arch-neutral algorithms wheels), the `wheels-py...` -> `algorithms-wheels-py...` rename has no other consumer in the repo's workflows, the `.cudaq_version` pin is byte-identical to the one cudaqx currently tests against, and the 0.15.99 stamps are consistent across the get-cudaq-wheels action, the test-side pin, and the pyproject bounds. The end-to-end validation run (30331620718, pre-0.15-migration) checks out all-green, and the current head's PR checks are green on all four legs. The fork story also strictly improves: the pure-Python build no longer needs the toolchain, and the test job is correctly skipped along with `build-cudaq-wheels`. Nice, well-documented matrix.

Five non-blocking inline comments: a now-dead `build_type` input; a comment-accuracy nit about `device` vs artifacts (and the cpu<->cu12 / gpu<->cu13 coverage split); a latent artifact-name collision if the algorithms build matrix ever regrows a platform axis; PyPI mode being unsatisfiable until a cuda-quantum 0.15.1 release exists; and a stale comment the new toggle left behind on `test-wheels`.

Two meta-nits: the PR title begins with a literal `# ` (a markdown header leaked into the title field — worth stripping so it doesn't become the subject of a squash commit), and the description predates the last two commits (it still says `scripts/ci/test_wheels.sh` is unchanged and cites the pre-0.15 validation run).

pyproject.toml.cu12
bmhowe23: Just checking - are you able to depend on the `cudaq` metapackage rather than depending on the CUDA-specific versions of CUDA-Q? That would disentangle one complexity for you: you'd be able to be CUDA-version agnostic and therefore just post one single wheel. Otherwise, you'll have to have cudaq-algorithms-cu{12,13} and a metapackage that autoselects between them. <-- It's not impossible to do this, but it's easier if you don't actually have to.

.github/workflows/wheel_algorithms.yaml
wsttiger: Done — dropped the input from both trigger blocks and the matching `build_type: Release` in `pr_workflow.yaml` in the same commit. Left the inert `--build-type` flag in `build_wheels.sh` as interface-compat alongside the other unused-but-accepted flags.

.github/workflows/wheel_algorithms.yaml
wsttiger: Reworded all three comments to state the pinning is deliberate: `device` selects runner/container/simulator and — as legs are pinned — the CUDA major (cpu→cu12, gpu→cu13), each per-major wheel feeding 2 of the 4 legs, and the cross pairings intentionally out of scope for a smoke matrix. Also fixed the "fall back" wording in `build_cudaq_wheels.yaml`.

.github/workflows/wheel_algorithms.yaml
wsttiger: Good catch — deleted the `platform` axis and hardcoded `linux-amd64-cpu8` so the job can't structurally regrow a second arch leg that would collide on the arch-neutral artifact name.

pyproject.toml.cu12
wsttiger: Relaxed the floor to `>= 0.15.0, < 0.16` in a follow-up commit — released cuda-quantum 0.15.0 is on the public indices, so PyPI mode is now dispatchable today rather than waiting on 0.15.1, while Custom mode's from-source 0.15.99 still satisfies the range. Also fixed the misleading "pypi.nvidia.com" wording in the toggle description.

.github/workflows/wheel_algorithms.yaml
wsttiger: Updated the comment to cover both modes — Custom (consumes from-source wheels, both skipped on forks) and PyPI (build-cudaq-wheels skipped, released CUDA-Q installed instead) — and called out that the guard is intentionally `!failure() && !cancelled()`, not `success()`, so it still runs when the upstream build was skipped.

pyproject.toml.cu12
wsttiger: Great call — and since cudaq-algorithms is pure Python, we're not tied to a specific CUDA-Q ABI, so we can depend on the `cudaq` metapackage and drop the cu{12,13} split plus our own selector entirely. I'll do exactly that: repoint the dependency at `cudaq` (>= 0.15.0, < 0.16) and collapse to a single CUDA-agnostic `cudaq-algorithms` py3-none-any wheel. I'm handling it as a dedicated follow-up rather than folding it into this matrix PR, so the packaging switch gets its own wheel-test run — in particular I want to confirm the custom/unreleased-CUDA-Q test path still resolves cleanly through `cudaq`. Thanks for flagging it.

wsttiger: Update — I ended up folding this into the PR rather than a follow-up. `cudaq-algorithms` now depends on the `cudaq` metapackage (>= 0.15.0, < 0.16) and ships a single `py3-none-any` wheel; the `cudaq-algorithms-cu{12,13}` split and our selector metapackage are gone (net −410 lines). Validated green across the arm64/x86 × CPU/GPU matrix in a Custom-mode dispatch — the wheel installs `--no-deps` against the from-source cuda-quantum and the suite passes on CPU (qpp-cpu) and GPU (nvidia-fp64). Thanks again for the nudge — it's a much cleaner surface.

ivanbasov: Post-merge follow-up on the "dispatchable today" part — one thing worth knowing before the first PyPI-mode dispatch: as merged, it will go red on both CPU legs, deterministically.

The collapsed wheel depends on `cudaq >= 0.15.0, < 0.16`, which resolves to the sdist-only `cudaq` 0.15.0 — and its install-time selector (`hatch_build.py`) falls back to **`cuda-quantum-cu13`** whenever CUDA detection fails. In the driverless, toolkit-less `ubuntu:24.04` CPU containers it always fails: the detection dlopens bare `libnvrtc`/`libcudart` sonames on the system linker path and then tries pynvml (which needs the driver's `libnvidia-ml`); the `cuda_toolkit[cudart]` pip shim can't help, since its libraries land under `site-packages/nvidia/...` — off the dlopen path — and the metadata hook runs inside pip's isolated build env anyway. `cuda-quantum-cu13==0.15.0` wheels exist for both arches, so the install *succeeds* with cu13, and the leg then dies at `pip show cuda-quantum-cu12` in `test_wheels.sh`. GPU legs are fine (cu13 expected, cu13 detected). One irony worth noting: the deleted in-repo selector fell back the other way ("Assuming CUDA 12.0"), so the collapse silently flipped the CPU-leg default from matching to mismatching.

Cheap fixes, pick one: assert per-mode reality on the CPU legs (`pip show cudaq` + an import check, or simply expect cu13 there); pre-install the intended `cuda-quantum-cu${cuda_major}==0.15.0` from the public index before the algorithms wheel, so the fallback never engages; or export `LD_LIBRARY_PATH` to the shim's lib directory so the detection genuinely sees CUDA 12 (which would also make the "Help CUDA detection" comment true). Whichever way — one PyPI-mode dispatch would be worth running: with Custom's `--no-deps` bypass, no run so far has exercised the shipped wheel's real dependency chain.

PR #24, opened 2026-07-28
Add PySCF/Psi4 integral extraction to the chemistry bridge

wsttiger:
# Add Psi4 integral extraction alongside the PySCF chemistry bridge

## What

Adds two library helpers to `cudaq_algorithms.chemistry`:

- `from_pyscf(mean_field)` — chemist `(pq|rs)` MO integrals + nuclear
  repulsion from a converged restricted PySCF mean field.
- `from_psi4(wavefunction)` — the same, from a restricted Psi4
  wavefunction.

Both return `(one_body, eri, nuclear_repulsion)` in the exact form
`chemistry.qubit_hamiltonian(...)` already consumes.

## Why it's small

The conversion itself (`spin_orbital_tensors` / `qubit_hamiltonian`) is
**package-agnostic** — it takes plain NumPy chemist integrals and knows
nothing about their origin. So "support Psi4" needs no new conversion
logic, only a second extraction helper that produces the same tensors.
The PySCF path was previously inlined in a test; this promotes both to
documented library functions with a shared contract.

## Tests (`tests/python/test_psi4_conversion.py`)

1. **`from_pyscf` reproduces FCI** for H2 and H4 (STO-3G) — machine
   precision (diff ~1e-15). Runs wherever PySCF is installed.
2. **PySCF vs Psi4 agree** — same molecule/basis extracted from both,
   qubit Hamiltonians compared by **spectrum**. Spectrum, not
   coefficient-by-coefficient equality, is the correct invariant: the
   two codes' canonical MOs can differ by orbital phase or (for
   degeneracies) ordering — a unitary gauge that leaves every physical
   quantity unchanged. Verified the comparison is gauge-robust (sign
   flips / orbital swaps → 0 change) yet bug-sensitive (wrong scaling,
   perturbation, or physicist-ordering → 0.1–0.9 change, far past the
   1e-6 tolerance).

**Executed and green.** All four tests pass with **Psi4 1.10 / PySCF
2.13.1** (installed via micromamba/conda-forge, run against CUDA-Q
v0.15): measured PySCF-vs-Psi4 spectrum agreement is ~8e-10 (H2) and
~3e-9 (H4). One subtlety surfaced and handled: the two codes use
marginally different Bohr-radius constants for the Å→bohr conversion, so
nuclear repulsion (and the integrals) differ at the ~1e-9 level — the
nuclear-repulsion sanity check is at 1e-6, and the spectrum tolerance
(1e-6) absorbs it with ~1000x margin.

## Notes for review

- **Psi4 availability.** Psi4 ships via conda-forge (no PyPI wheel), so
  the cross-check test is `importorskip`-guarded. I ran it locally in a
  micromamba env (Psi4 1.10) — all four tests pass — but **CI has no
  conda-enabled job**, so in the current CI the Psi4 cross-check will
  *skip*, not run. Adding a conda-based test job (or a psi4 wheel source)
  is a follow-up if we want it enforced in CI.


ivanbasov (commented):
Reviewed both files and, since the two engines are runnable, verified the substance by execution rather than by reading alone: the FCI test reproduces (agreement ~1e-15, as the description claims), the PySCF-vs-Psi4 spectrum agreement reproduces at ~1e-9 (Psi4 1.11 / PySCF 2.14), the chemist-ordering conventions check out on both paths (`mints.mo_eri` really is chemist `(pq|rs)`; `ao2mo.restore(1, ...)` unpacks correctly against a dense `compact=False` reference), and the description's gauge-robustness / bug-sensitivity claims for the spectrum comparison reproduce as stated when measured on the full spectrum. Nice clean pair of helpers with a well-argued test design.

Five inline comments: one real correctness gap (hcore built as kin+nuc drops ECP terms — silently wrong integrals for ECP molecules, executed repro + one-line fix), a crash on the docstring's own default-usage path for symmetric molecules in Psi4, a CI-coverage gap (all four new tests skip in CI, not just the Psi4 half), unenforced restricted-only contracts, and a comment nit about s4-vs-s8 packing.

python/cudaq_algorithms/chemistry.py
```diff
@@ -123,3 +125,68 @@ def qubit_hamiltonian(one_body: ArrayLike,
...
+    """
+    from functools import reduce
+
+    from pyscf import ao2mo
+
+    mol = mean_field.mol
+    coefficients = np.asarray(mean_field.mo_coeff)
+    num_orbitals = coefficients.shape[1]
+
+    core_ao = mol.intor("int1e_kin") + mol.intor("int1e_nuc")
```
ivanbasov: Worth a fix before merge: building hcore by hand as `int1e_kin + int1e_nuc` drops everything `mean_field.get_hcore()` would add — most notably the `ECPscalar` term whenever the molecule carries an effective core potential (also X2C decoration, QM/MM charges, ...). An ECP-RHF is squarely inside the documented contract ("a converged restricted mean field, e.g. `pyscf.scf.RHF(mol).run()`"), and the failure is silent.

Reproduced at this head: NaH with `basis="lanl2dz", ecp="lanl2dz"`, plain `scf.RHF(mol).run()` — `max|kin+nuc - mf.get_hcore()| = 0.0774` (exactly `max|ECPscalar|`); the HF energy rebuilt from the returned `(one_body, eri, e_nuc)` is off by 3.6e-2 Ha (~23 kcal/mol) and FCI by 4.3e-2 Ha, with no error or warning. Swapping in `core_ao = mean_field.get_hcore()` reproduces `mf.e_tot` exactly, and is identical for the all-electron case (verified on H2/H4).

A side effect worth knowing: Psi4's `wavefunction.H()` *does* include the ECP (verified — same 0.0774 shows up there), so as written the "identical in meaning and convention to :func:`from_pyscf`" contract in `from_psi4`'s docstring is false precisely for ECP systems: the two helpers return Hamiltonians differing by ~36 mHa for the same molecule. `get_hcore()` fixes that too. Might also be worth adding one ECP molecule to `_MOLECULES` so this input class stays pinned.

python/cudaq_algorithms/chemistry.py
```diff
@@ -123,3 +125,68 @@ def qubit_hamiltonian(one_body: ArrayLike,
...
+    :func:`from_pyscf`, so either drives :func:`qubit_hamiltonian`
+    unchanged.
+
+    Restricted references only (uses ``Ca``); ``mo_eri`` already returns
+    the chemist ``(pq|rs)`` ordering, matching the PySCF path.
+    """
+    import psi4
+
+    coefficients_matrix = wavefunction.Ca()
+    coefficients = np.asarray(coefficients_matrix)
```
ivanbasov, line 205: Psi4 auto-detects the point group by default, so the docstring's own usage example — "the second return value of `psi4.energy("scf", return_wfn=True)`" — crashes here for almost any symmetric molecule (H2, N2, H2O, ...): `Ca()` comes back irrep-blocked and `np.asarray` raises `ValidationError: __array__interface__ can only be called on Psi4 data object with only one irrep!`, which names the condition but not the fix. The test file quietly sidesteps this by hardcoding `"symmetry c1"` into the geometry. (Reproduced on Psi4 1.11: default H2 -> d2h, nirrep 8 -> crash; with `symmetry c1` -> works.)

Suggestion: check `wavefunction.nirrep() != 1` and raise with an actionable message ("compute the wavefunction with symmetry c1"), or handle blocked matrices (`wavefunction.c1_deep_copy(...)`, or `Ca_subset("AO", "ALL")` plus a matching AO-basis core Hamiltonian) — and either way state the C1 requirement in the docstring.

tests/python/test_psi4_conversion.py
```diff
@@ -0,0 +1,108 @@
...
+
+def _spectrum(one_body, eri, nuclear_repulsion):
+    operator = algorithms.chemistry.qubit_hamiltonian(
+        one_body, eri, scalar_offset=nuclear_repulsion, tolerance=1e-12)
+    return np.linalg.eigvalsh(np.asarray(operator.to_matrix()))
+
+
+@pytest.mark.parametrize("label, basis, geometry", _MOLECULES)
+def test_from_pyscf_reproduces_fci(label, basis, geometry):
+    pytest.importorskip("pyscf")
```
ivanbasov, line 76: CI-coverage gap, and wider than the "Notes for review" describe: neither CI pytest entry point installs pyscf (`scripts/ci/test_wheels.sh` installs only `pytest scipy`; the Algorithms job installs `numpy scipy pytest`), so it isn't just the Psi4 cross-check that skips — *all four* tests in this file skip in CI, including the FCI check, which is the only independent end-to-end pin of the new helpers (verified: `4 skipped`, suite green, under the CI dependency set). The two new public functions land with zero enforced coverage — the same failure mode as the round-1 skip-guard finding on the double-factorization bridge.

pyscf ships manylinux wheels for both x86_64 and aarch64, and the two FCI tests take ~11 s, so adding `pyscf` to the CI test installs would make the PySCF leg real. (Alternative: freeze the H2/H4 reference integrals + FCI energies as constants, the `test_fermion_compilers.py` pattern.) Worth also touching up the PR description so the skip story covers both engines.

python/cudaq_algorithms/chemistry.py
```diff
@@ -123,3 +125,68 @@ def qubit_hamiltonian(one_body: ArrayLike,
...
+
+    Restricted (single ``mo_coeff`` matrix) references only; the spin
+    expansion downstream assumes one spatial set shared by both spins.
+    """
+    from functools import reduce
+
+    from pyscf import ao2mo
+
+    mol = mean_field.mol
+    coefficients = np.asarray(mean_field.mo_coeff)
```
ivanbasov, line 151: Non-blocking: "Restricted ... references only" is documented but unenforced, and the failure modes never mention the contract. A PySCF UHF mean field (`mo_coeff` shape `(2, nao, nmo)`) dies in the `reduce(np.dot, ...)` below with a bare numpy shape error — and in the `nao == 2` corner (H2/sto-3g) it actually gets all the way to `qubit_hamiltonian`, which then complains about a non-square `one_body` that the library itself built. On the Psi4 side a UHF wavefunction silently uses `Ca()` — alpha orbitals only.

Cheap guards with actionable messages: here, `np.asarray(mean_field.mo_coeff).ndim != 2 -> raise` ("restricted mean fields only"); in `from_psi4`, `not wavefunction.same_a_b_orbs() -> raise`.

python/cudaq_algorithms/chemistry.py
```diff
@@ -123,3 +125,68 @@ def qubit_hamiltonian(one_body: ArrayLike,
...
+    num_orbitals = coefficients.shape[1]
+
+    core_ao = mol.intor("int1e_kin") + mol.intor("int1e_nuc")
+    one_body = reduce(np.dot, (coefficients.T, core_ao, coefficients))
+
+    # ao2mo.full returns chemist-notation MO integrals; restore(1, ...)
+    # unpacks the 8-fold-symmetric storage to the dense (n, n, n, n) tensor.
```
ivanbasov: Comment nit: `ao2mo.full(mol, C)` with the default `compact=True` returns the 4-fold (s4) packed matrix of shape `(npair, npair)` — pyscf documents `full` as guaranteeing at most 4-fold symmetry — not "the 8-fold-symmetric storage". `restore(1, ...)` auto-detects the layout, so the code is right (verified against the `compact=False` dense reference at ~9e-16); the comment just documents a storage format that never occurs here.

python/cudaq_algorithms/chemistry.py
wsttiger: Fixed in `50beae7` — now uses `mean_field.get_hcore()`, which carries ECPscalar/X2C/QM-MM. All-electron H2/H4 are unchanged (get_hcore is identical there), so the existing tests still pass. On pinning an ECP molecule: the test's dense `qubit_hamiltonian(...).to_matrix()` spectrum is 4^(2n), and a real ECP system (NaH/lanl2dz) has far too many orbitals for that — it OOMs. I left `_MOLECULES` as-is rather than add a classical-only HF-energy check in a different style; happy to add one if you'd prefer the path pinned. Note the same `get_hcore()` change also restores the `from_psi4`/`from_pyscf` convention match you flagged — Psi4's `H()` already includes the ECP, so both now agree for ECP systems.

python/cudaq_algorithms/chemistry.py
wsttiger: Fixed in `50beae7` — `from_psi4` now raises a ValueError naming the fix ("Recompute with symmetry c1 …") when `nirrep() != 1`, and the docstring states the C1 requirement. Verified default-symmetry H2 (nirrep 8) hits the guard cleanly.

tests/python/test_psi4_conversion.py
wsttiger: Fixed in `50beae7` — added `pyscf` to both CI test-dependency installs (`scripts/ci/test_wheels.sh` and the Algorithms job), so the two `from_pyscf` FCI tests now run instead of skip. The Psi4 leg still skips (conda-forge only, no PyPI wheel). I'll also touch up the PR description's skip story to cover both engines.

python/cudaq_algorithms/chemistry.py
wsttiger: Fixed in `50beae7` — `from_pyscf` guards `mo_coeff.ndim != 2` ("restricted mean fields only"), and `from_psi4` guards `not wavefunction.same_a_b_orbs()` ("restricted reference only"). Both raise before the downstream shape error. Verified a UHF pyscf input now hits the guard.

python/cudaq_algorithms/chemistry.py
wsttiger: Fixed in `50beae7` — comment now says "at most 4-fold packed storage" rather than 8-fold. `restore(1, …)` auto-detects either way, so no code change.

ivanbasov (approved):
Round 2 — verified `50beae7` by execution; all five comments delivered:

- **ECP hcore**: re-ran the round-1 repro (NaH, `basis=ecp="lanl2dz"`, plain RHF) — the HF energy rebuilt from the returned tensors now matches `mf.e_tot` exactly (was off by 3.6e-2 Ha), and the same `get_hcore()` change restores the `from_pyscf`/`from_psi4` convention match for ECP systems.
- **Guards**: a PySCF UHF mean field now raises the restricted-contract message at entry; default-symmetry H2 in Psi4 (nirrep 8) raises the "Recompute with 'symmetry c1'" error; a UHF Psi4 wavefunction hits the `same_a_b_orbs` guard (all three exercised live on Psi4 1.11 / PySCF 2.14). The RHF + C1 happy path is unchanged.
- **CI**: with `pyscf` in both install lines, the file runs as 2 passed / 2 skipped (psi4 only) under the CI dependency set — the FCI leg is now enforced rather than silently skipped.
- Comment/docstring rewordings check out (s4 packing; C1 requirement).

Answered the ECP-pinning question in its thread — non-blocking either way.

One logistics heads-up: #23's merge rewrote `scripts/ci/test_wheels.sh` for the single-wheel packaging, so this branch now conflicts there; the `pyscf` addition will need a small rebase onto the new Custom/PyPI split (the shared test-deps install is line 27 in the merged version).

python/cudaq_algorithms/chemistry.py
ivanbasov: Confirmed by re-running the round-1 repro at `50beae7`: the rebuilt HF energy now matches `mf.e_tot` exactly (0.0 at 1e-12 print precision; was 3.63e-2 Ha), and the `get_hcore()` change indeed restores the `from_pyscf`/`from_psi4` agreement for ECP molecules.

On the pinning question: your classical HF-energy check is exactly the right shape, and it's non-blocking from my side. No qubit map needed — rebuild the closed-shell energy from the returned triple (`2·Σ h_ii + Σ_ij [2(ii|jj) − (ij|ji)]` over occupied spatial orbitals, plus `e_nuc`) and compare to `mf.e_tot` at ~1e-10 for one ECP molecule; NaH/lanl2dz converges in well under a second. That's the exact check I used to verify this fix, and it pins precisely the input class the original comment was about with no `to_matrix()` blow-up. Happy with it as a test here or as a follow-up.

PR #25, opened 2026-07-28
Add an FCIDUMP reader to the chemistry bridge

wsttiger:
## What

Adds `cudaq_algorithms.chemistry.from_fcidump(contents)` — a reader for
the FCIDUMP molecular-integral interchange format (written by Molpro,
PySCF, Psi4, Block/DMRG, …). It returns the same
`(one_body, eri, core_energy)` triple as `from_pyscf`/`from_psi4`, in
chemist `(pq|rs)` notation over real spatial orbitals — the exact form
`chemistry.qubit_hamiltonian(...)` and `DoubleFactorizedEncoding(...)`
already consume. A `.fcidump` file thus becomes a one-line front end to
the whole pipeline.

## Why it's small (and dependency-free)

FCIDUMP stores chemist-ordered `(ij|kl)` integrals over real spatial
orbitals — precisely the convention the bridge already speaks. So the
reader is a direct 1-based → 0-based copy with the eight-fold index
symmetry filled from the symmetry-unique records; no transpose, no
reordering. It's pure text + NumPy — **no third-party dependency**, so it
works in the pure-Python wheel with no PySCF/Psi4 install. 

## Tests (`tests/python/test_fcidump.py`)

All primary tests inject a string constant — no files, no PySCF:

1. **Reproduces reference integrals** — parses an H2/STO-3G FCIDUMP and
   matches the literature `H1`/`ERI`/`E_NUCLEAR` (the same values
   `test_df_qsvt_bridge.py` uses) to 1e-12.
2. **Eight-fold symmetry fill** — a single `(21|21)` record populates all
   four orbit elements, and the parsed `eri` passes
   `spin_orbital_tensors`' chemist-symmetry check.
3. **Spectrum equality** — the qubit Hamiltonian built from the parsed
   tensors matches the one from the integrals fed directly (1e-10), and
   recovers the H2 FCI energy — the convention-level check.
4. **Tolerant parsing** — a variant with lowercase `&fci`, a `/`
   namelist terminator, Fortran `D` exponents, comments, blank lines, and
   irregular spacing parses to the identical tensors.
5. **Validation errors** — missing header, missing `NORB`, a malformed
   line, and an out-of-range index each raise a specific `ValueError`.
6. **PySCF interop (bonus, `importorskip`)** — round-trips through
   PySCF's canonical `fcidump.from_integrals` writer and compares spectra.

**Executed and green** against CUDA-Q v0.15 (9 passed, incl. the PySCF
round-trip on PySCF 2.13.1); full suite stays green.

## Scope / notes for review

- **Real (RHF/ROHF) FCIDUMP only.** Complex/UHF variants (`IUHF=1`) have a
  different index symmetry; they're out of scope and are rejected
  downstream by the existing `validate_symmetry` check. Noted in the
  docstring.
- **`NELEC`/`MS2` are parsed-region metadata we don't return** (only
  `NORB` is required, to size the tensors). If state-prep later needs the
  electron count, a small `read_fcidump_header` can expose it without
  touching this 3-tuple signature.

### DIFF
--- a/docs/double_factorization.md
+++ b/docs/double_factorization.md
@@ -161,3 +161,29 @@ factorization itself does not. See
 which sweeps X-DF leaf counts on H2/STO-3G (hardcoded integrals, no PySCF
 needed) and tabulates tensor error, `alpha`, term count, and the exact
 ground-state shift at each truncation.
+
+### Loading integrals
+
+Integrals can come from a mean-field object (`chemistry.from_pyscf` /
+`chemistry.from_psi4`) or from an **FCIDUMP** file — the de-facto
+interchange format written by Molpro, PySCF, Psi4, and Block/DMRG codes.
+`chemistry.from_fcidump` parses the *text* of such a file (the caller does
+the file I/O, so the parse stays pure and testable) and returns the same
+`(one_body, eri, core_energy)` triple the other loaders do, already in
+chemist `(pq|rs)` notation over real spatial orbitals:
+
+```python
+from pathlib import Path
+from cudaq_algorithms import chemistry
+
+one_body, eri, core = chemistry.from_fcidump(
+    Path("molecule.fcidump").read_text())
+h = chemistry.qubit_hamiltonian(one_body, eri, scalar_offset=core)
+# or hand the same tensors to DoubleFactorizedEncoding(one_body, eri, ...)
+```
+
+The reader has no third-party dependency (pure text + NumPy) and fills the
+eight-fold index symmetry from the symmetry-unique records FCIDUMP stores.
+Only real (RHF/ROHF) FCIDUMP files are supported; complex/UHF variants
+(`IUHF=1`) carry a different index symmetry and are rejected by the
+`validate_symmetry` check downstream.
--- a/docs/double_factorization.md
+++ b/docs/double_factorization.md
@@ -164,13 +164,13 @@ ground-state shift at each truncation.
 
 ### Loading integrals
 
-Integrals can come from a mean-field object (`chemistry.from_pyscf` /
-`chemistry.from_psi4`) or from an **FCIDUMP** file — the de-facto
-interchange format written by Molpro, PySCF, Psi4, and Block/DMRG codes.
+Integrals can come from an **FCIDUMP** file — the de-facto interchange
+format written by Molpro, PySCF, Psi4, and Block/DMRG codes.
 `chemistry.from_fcidump` parses the *text* of such a file (the caller does
-the file I/O, so the parse stays pure and testable) and returns the same
-`(one_body, eri, core_energy)` triple the other loaders do, already in
-chemist `(pq|rs)` notation over real spatial orbitals:
+the file I/O, so the parse stays pure and testable) and returns the
+`(one_body, eri, core_energy)` triple `qubit_hamiltonian` and
+`DoubleFactorizedEncoding` consume, already in chemist `(pq|rs)` notation
+over real spatial orbitals:
 
 ```python
 from pathlib import Path
@@ -184,6 +184,7 @@ h = chemistry.qubit_hamiltonian(one_body, eri, scalar_offset=core)
 
 The reader has no third-party dependency (pure text + NumPy) and fills the
 eight-fold index symmetry from the symmetry-unique records FCIDUMP stores.
-Only real (RHF/ROHF) FCIDUMP files are supported; complex/UHF variants
-(`IUHF=1`) carry a different index symmetry and are rejected by the
-`validate_symmetry` check downstream.
+Only real (RHF/ROHF) FCIDUMP files are supported; unrestricted variants
+(Molpro's `IUHF=1` or Psi4's `UHF=.TRUE.`) carry a spin-resolved integral
+set with a different index symmetry and are rejected up front by a header
+guard.
--- a/python/cudaq_algorithms/chemistry.py
+++ b/python/cudaq_algorithms/chemistry.py
@@ -22,10 +22,12 @@ called, not at import.
 
 from __future__ import annotations
 
+import re
+
 import numpy as np
 from numpy.typing import ArrayLike
 
-__all__ = ["spin_orbital_tensors", "qubit_hamiltonian"]
+__all__ = ["from_fcidump", "spin_orbital_tensors", "qubit_hamiltonian"]
 
 
 def spin_orbital_tensors(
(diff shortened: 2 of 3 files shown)


ivanbasov (commented):
Reviewed the parser, tests, and docs, and verified the substance by execution: the strict and tolerant fixtures parse to the reference integrals, the spectrum path reproduces the H2 FCI energy, and round-trips through the real PySCF writer (`fcidump.from_integrals` and `from_scf`, including a symmetry-adapted molecule with nontrivial ORBSYM) come back spectrum-exact. Also nice to see most of this test file ungated — unlike its sibling PR, these tests genuinely run in CI. Clean, dependency-free reader.

Four inline comments, all non-blocking in tone but the first one deserves a look before merge: the "UHF/IUHF=1 files are rejected downstream by validate_symmetry" story is provably false (the parser's own symmetry fill makes its output pass that check *by construction* — real UHF files parse silently into wrong tensors; a one-line header guard fixes it); the parser rejects standard orbital-energy records (`value i 0 0 0`) that most Molpro-lineage FCIDUMPs contain; the new docs section references `chemistry.from_pyscf`/`from_psi4`, which don't exist on this branch (they land in the sibling PR); and the test docstring cites a `docs/conventions.md` that isn't in the repo.

python/cudaq_algorithms/chemistry.py
```diff
@@ -123,3 +125,115 @@ def qubit_hamiltonian(one_body: ArrayLike,
...
+        hamiltonian = qubit_hamiltonian(one_body, eri, scalar_offset=core)
+
+    FCIDUMP indices are Fortran 1-based; a ``value i j k l`` record with all
+    of ``i,j,k,l`` nonzero is a two-electron integral, with ``k == l == 0``
+    a one-electron integral ``h_ij`` (its transpose is filled too), and with
+    all indices zero the core energy.
+
+    Only real (RHF/ROHF-style) FCIDUMP files are supported; complex/UHF
+    variants (``IUHF=1``) have a different index symmetry and are rejected
+    downstream by the ``validate_symmetry`` check in ``qubit_hamiltonian``.
```
ivanbasov: Worth a fix before merge: this sentence (same story in `docs/double_factorization.md` and the PR description) promises that complex/UHF variants "are rejected downstream by the `validate_symmetry` check" — but that check can never fire on `from_fcidump` output, for a structural reason: every two-electron record is written onto its full 8-fold orbit (and orbits partition the index space), and every one-electron record fills both triangles, so the returned tensors satisfy exactly the symmetries `validate_symmetry` checks *by construction*, whatever the input file was.

Demonstrated at this head with two real inputs: a Knowles-Handy `IUHF=1` file (aa/bb/ab blocks with `0.0 0 0 0 0` delimiters) parses without a murmur — the beta blocks overwrite the alpha values in the same spatial slots (`one_body` diag comes back `[-1.2, -0.5]` where alpha was `[-1.1, -0.4]`; `eri[0,0,0,0]` ends up as the ab-block value), the delimiter records are consumed as core-energy records — and `spin_orbital_tensors(..., validate_symmetry=True)` passes. A genuine Psi4-written `UHF=.TRUE.` dump behaves the same. So instead of the promised rejection, a user gets a silently wrong Hamiltonian.

Since the header is already parsed, the guard is one line: reject `IUHF\s*=\s*[1-9]` (and Psi4's `UHF=.TRUE.` spelling; `\bUHF` won't false-match inside `IUHF`) with a clear "unrestricted FCIDUMP files are not supported" — and reword this sentence and the docs paragraph accordingly.

python/cudaq_algorithms/chemistry.py
```diff
@@ -123,3 +125,115 @@ def qubit_hamiltonian(one_body: ArrayLike,
...
+                eri[a, b, c, d] = value
+        elif i and j and not k and not l:
+            one_body[i - 1, j - 1] = value
+            one_body[j - 1, i - 1] = value
+        elif not (i or j or k or l):
+            core_energy = value
+        else:
+            raise ValueError(
+                f"unexpected FCIDUMP index pattern (a one-electron record "
+                f"must have k = l = 0): {line!r}")
```
ivanbasov: This `else` also catches a *standard* record type: orbital energies, `value i 0 0 0` (optional in the Knowles-Handy format). They're common in the wild — 173 of the 242 FCIDUMP files in NECI's test corpus carry them — and Psi4 emits them under its documented HANDE workflow (`psi4.fcidump(wfn, oe_ints=['EIGENVALUES'])`); pyscf's own reader tolerates them. Here they raise, and with a self-contradictory message: "a one-electron record must have k = l = 0" — for a record that *does* have k = l = 0 (it's `j` that's zero).

Suggestion: add an `elif i and not (j or k or l):` branch that skips (or collects) orbital-energy records, keep the raise for genuinely malformed patterns like `(i, 0, k, l)`, and drop one such record into the tolerant-parsing fixture so the behavior is pinned by the CI-run tests.

docs/double_factorization.md
```diff
@@ -161,3 +161,29 @@ factorization itself does not. See
 which sweeps X-DF leaf counts on H2/STO-3G (hardcoded integrals, no PySCF
 needed) and tabulates tensor error, `alpha`, term count, and the exact
 ground-state shift at each truncation.
+
+### Loading integrals
+
+Integrals can come from a mean-field object (`chemistry.from_pyscf` /
```
ivanbasov: Cross-branch reference: this section (and the `from_fcidump` docstring's "the same triple ``from_pyscf``/``from_psi4`` return") points users at `chemistry.from_pyscf` / `chemistry.from_psi4`, but neither exists on this branch — they're added by the sibling PR (#24), which is unmerged and independent (both branch from main). If this PR lands first, the docs direct users to an `AttributeError`.

Either declare/stack the dependency, or word it branch-neutrally (e.g. "the `(one_body, eri, core_energy)` triple in chemist notation that `qubit_hamiltonian` and `DoubleFactorizedEncoding` consume") until #24 lands.

tests/python/test_fcidump.py
```diff
@@ -0,0 +1,164 @@
...
+# the terms of the Apache License 2.0 which accompanies this distribution.     #
+# ============================================================================ #
+"""FCIDUMP reader -> chemist-notation tensors -> qubit Hamiltonian.
+
+The integral data is injected as an in-line string (the contents a caller
+would read from a `.fcidump` file), so these tests need no external files.
+The parsed tensors are pinned against the same literature H2/STO-3G values
+`test_df_qsvt_bridge.py` uses, and the resulting qubit Hamiltonian is
+cross-checked by whole-spectrum equality against those integrals fed
+directly -- the convention-level check `docs/conventions.md` recommends.
```
ivanbasov: Tiny one: `docs/conventions.md` doesn't exist in the repo (docs/ has df_block_encoding, double_factorization, fermion_transforms, pauli_lcu_qsvt, stateprep, trotter). Point this at the intended doc or drop the reference.

python/cudaq_algorithms/chemistry.py
wsttiger: Good catch, and you're right about the mechanism — the 8-fold fill makes the output pass `validate_symmetry` by construction, so the old sentence described a rejection that can never fire. Replaced it with an explicit header guard in `ba6a8b4`: `from_fcidump` now rejects `IUHF=1` / `UHF=.TRUE.` up front with a clear message, and the docstring/docs are reworded to match. Added tests pinning both spellings, plus an `IUHF=0` case so the guard doesn't false-match restricted files.

python/cudaq_algorithms/chemistry.py
wsttiger: Fixed in `ba6a8b4`. Added an `elif i and not (j or k or l):` branch that skips orbital-energy records, kept the raise for genuinely malformed patterns (and dropped the self-contradictory wording), and pinned it in CI: the tolerant fixture now carries two `value i 0 0 0` records and a new test asserts they're skipped, leaving the tensors identical.

docs/double_factorization.md
wsttiger: Reworded branch-neutrally in `ba6a8b4` — both the docs section and the docstring line you flagged now describe the `(one_body, eri, core_energy)` triple that `qubit_hamiltonian`/`DoubleFactorizedEncoding` consume, with no reference to the sibling-PR loaders. (Heads-up: #24 and this PR both add a `from_*` helper and an `__all__` entry to `chemistry.py`; they union-merge cleanly, just a merge-order note.)

tests/python/test_fcidump.py
wsttiger: Fixed in `ba6a8b4` — repointed to `docs/double_factorization.md`, which is where the chemist-notation convention actually lives.

ivanbasov (approved):
Round 2 — verified `ba6a8b4` by execution; all four delivered:

- **UHF guard**: a Knowles-Handy `IUHF=1` file (aa/bb blocks with zero-record delimiters) and a Psi4-style `UHF=.TRUE.` header both now fail fast with the clear "unrestricted" error; `IUHF=0` and `UHF=.FALSE.` don't false-trip — the latter checked against real Psi4 dumps, which always carry `UHF=.FALSE.` in the header — and lowercase `uhf=.true.` trips as intended. The docstring and docs no longer promise the (unfireable) downstream rejection.
- **Orbital-energy records**: a fresh `psi4.fcidump(..., oe_ints=['EIGENVALUES'])` dump now parses, with tensors bit-identical to the same dump without those records; the tolerant fixture + new test pin the behavior in CI.
- Full suite: 13 passed / 0 skipped in a pyscf env (the writer round-trip included), and the docs/docstring are branch-neutral with the conventions reference repointed.

LGTM — thanks for the thorough turnaround, the `IUHF=0` no-false-match test included.

PR #26, opened 2026-08-03
Move double factorization encoding to examples

wsttiger:
Move the double-factorized encoding to a bring-your-own-encoding example, on real molecules

## Description

`DoubleFactorizedEncoding` leaves the package surface and becomes
[`examples/bring_your_own_encoding/df_encoding.py`](examples/bring_your_own_encoding/df_encoding.py):
a complete, runnable demonstration that the `BlockEncoding` protocol is
structural — implement the protocol's surface against the public API and
`Walk`/`QSVT` accept the encoding unchanged. Its companion demo drives the
encoding head to head with the built-in `PauliLCU` on **real molecules**
(integrals built live from PySCF), across a six-configuration menu.

**Motivation.** The frame-cancellation construction at the heart of this
encoding relies on a shared orthogonal frame per leaf, which is specific to
the double-factorized form and does not generalize to the THC family the
library is converging toward. Rather than carry two DF encodings on the API
long-term, this repositions the original as the extension-protocol
showcase, preserving its pedagogical value and its full validation suite.

### What changed

- `python/cudaq_algorithms/df_encoding.py` →
  `examples/bring_your_own_encoding/df_encoding.py`. The example depends
  only on the public package surface: relative imports become absolute, and
  the two small private helpers it used (`_validate_power`,
  `_prepare_angles`) are inlined, so the file can be copied wholesale as
  the starting point for a new encoding.
- `cudaq_algorithms.DoubleFactorizedEncoding` and the `df_encoding`
  submodule are removed from the package root (pointer comment left in
  `__init__.py`). The classical `double_factorization` preprocessing
  package is untouched and remains public.
- The companion demo (`df_block_encoding.py`, formerly under
  `examples/double_factorization/`) moves beside the encoding and is
  rebuilt around a **PySCF molecule menu** (`pip install pyscf`):

  ```
  h2         H2 / STO-3G           2 orbitals ->  4 system qubits  (default)
  h2o-cas44  H2O / STO-3G CAS(4,4) 4 orbitals ->  8 system qubits
  h4         linear H4 / STO-3G    4 orbitals ->  8 system qubits
  lih        LiH / STO-3G          6 orbitals -> 12 system qubits
  h2o        H2O / STO-3G          7 orbitals -> 14 system qubits
  h2o-631g   H2O / 6-31G          13 orbitals -> 26 system qubits
  ```

  Every configuration tells the classical story: DF vs flat-`PauliLCU`
  alpha and term counts, the leaf-truncation dial, and an **RC-DF row**
  (re-optimize the kept leaves at the same budgets — complementing the
  classical C-DF-vs-X-DF comparison in `examples/double_factorization/`
  with the encoding-side view, and the first *regularized* (RC-DF) use
  in an example). Small
  configurations also run the circuits, self-verified against a sparse
  Jordan–Wigner reference (~1e-15) with Chebyshev moments through the
  shared `Walk`; larger ones print the statevector cost and skip execution.
  The active-space entry runs the **fully verified circuit story on real
  water** via PySCF's CASCI effective integrals. At H2O/6-31G the DF alpha
  is 145.1 vs 212.3 for the flat expansion (~32% lower).
- Docs updated (`docs/df_block_encoding.md` — retitled, menu documented;
  `docs/double_factorization.md`; a `chemistry.py` docstring mention).

### Compatibility notes

- **API removal** (pre-1.0): `from cudaq_algorithms import
  DoubleFactorizedEncoding` no longer works; the class is imported from the
  example file instead (snippet in `docs/df_block_encoding.md`).
- The wheel loses the `df_encoding` module (that is the disclosed API
  removal); nothing else in the artifact changes — examples and tests were
  never packaged, and nothing remaining in the shipped package depended on
  it.
- PySCF is an **example-only** dependency (friendly import error points to
  `pip install pyscf`); the test suite does not import the demo, so CI
  needs no new dependency.

### Validation

- Full suite on the branch: **269 passed, 3 skipped** (fp64 `qpp-cpu`);
  every touched file clean under the CI-pinned yapf 0.43.0.
- All six demo configurations run green with their self-checks
  (h2 4 s; h2o-cas44 30 s; h4 ~2 min; lih 23 s; h2o 42 s; h2o-631g 3 s).
- Two independent review rounds on the move itself; all findings addressed
  (the CI paths-filter gap and the helper divergence pins above).

### DIFF
--- a/docs/df_block_encoding.md
+++ b/docs/df_block_encoding.md
@@ -78,8 +78,10 @@ construction it reproduces the LCU one-norm of
 alpha = |const| + sum_k |F_k| + sum_t ( sum_{k<l} |Z^t_kl| + 1/4 sum_k |Z^t_kk| )
 ```
 
-Compressing the factorization (fewer leaves) lowers `alpha` and the term
-count together — the knob a flat Pauli expansion does not have. Since
+Compressing the factorization (fewer leaves) shrinks the term count, and
+with it (typically, though not monotonically — dropping a leaf also
+reshapes the one-body singles absorbed into `kappa`) the `alpha` — the
+knob a flat Pauli expansion does not have. Since
 QSVT circuit depth for time evolution scales like `alpha * t`, the
 compression translates directly into shallower circuits, at the price of
 a spectrum shift bounded by the tensor reconstruction error.
@@ -108,7 +110,7 @@ runs the encoding on real molecules (integrals from PySCF —
 `pip install pyscf`), across a menu of configurations:
 
 ```
-python3 df_block_encoding.py [config]
+python3 df_block_encoding.py [config] [--circuits]
 
 h2         H2 / STO-3G           2 orbitals ->  4 system qubits  (default)
 h2o-cas44  H2O / STO-3G CAS(4,4) 4 orbitals ->  8 system qubits
--- a/docs/df_block_encoding.md
+++ b/docs/df_block_encoding.md
@@ -16,6 +16,8 @@ dense-reference test suite (`tests/python/test_df_encoding.py`) stays in
 CI, so the example is held to library-grade correctness.
 
 ```python
+import sys
+sys.path.insert(0, "examples/bring_your_own_encoding")  # from the repo root, or copy the file
 from df_encoding import DoubleFactorizedEncoding   # the example module
 from cudaq_algorithms import Walk, QSVT
 from cudaq_algorithms import double_factorization as df
--- a/docs/df_block_encoding.md
+++ b/docs/df_block_encoding.md
@@ -7,12 +7,12 @@ integrals (von Burg et al., *PRX Quantum* **2**, 030305 (2021);
 expanding it into Pauli words. It satisfies the same `BlockEncoding`
 protocol as `PauliLCU`, so `Walk` and `QSVT` consume it unchanged.
 
-It is intentionally **not part of the package**: it lives at
+It ships as a runnable example —
 [`examples/bring_your_own_encoding/df_encoding.py`](../examples/bring_your_own_encoding/df_encoding.py)
-as the worked, full-scale demonstration that the `BlockEncoding` protocol
-is structural — a complete encoding implemented outside the library,
-against the public API only, plugs into every consumer. Its
-dense-reference test suite (`tests/python/test_df_encoding.py`) stays in
+— and doubles as the worked, full-scale demonstration that the
+`BlockEncoding` protocol is structural: implement the protocol's surface
+against the public API and every consumer accepts the encoding. Its
+dense-reference test suite (`tests/python/test_df_encoding.py`) runs in
 CI, so the example is held to library-grade correctness.
 
 ```python
--- a/docs/df_block_encoding.md
+++ b/docs/df_block_encoding.md
@@ -121,11 +121,11 @@ h2o-631g   H2O / 6-31G          13 orbitals -> 26 system qubits
 ```
 
 Every configuration compares `DoubleFactorizedEncoding` with a `PauliLCU`
-of the same Hamiltonian (alpha, term count, structure), sweeps the
-factorization-truncation dial, and re-optimizes the kept leaves with
-RC-DF at the same budgets (the second dial: optimize, don't just
-truncate -- with a small ridge so the optimizer cannot trade a huge
-one-norm for fit). Small configurations additionally verify
+of the same Hamiltonian (alpha, term count, structure) and sweeps the
+factorization-truncation dial; all but the largest also re-optimize the
+kept leaves with RC-DF at the same budgets (the second dial: optimize,
+don't just truncate -- with a small ridge so the optimizer cannot trade a
+huge one-norm for fit). Small configurations additionally verify
 the encoded block against a sparse Jordan-Wigner reference and measure
 Chebyshev moments through the shared `Walk` consumer; large ones report
 the statevector cost and tell the classical scaling story instead (at
--- a/docs/df_block_encoding.md
+++ b/docs/df_block_encoding.md
@@ -104,7 +104,24 @@ where the qubits are Z positions *in that frame's rotated basis*).
 ## Example
 
 [`examples/bring_your_own_encoding/df_block_encoding.py`](../examples/bring_your_own_encoding/df_block_encoding.py)
-builds a small system, compares `DoubleFactorizedEncoding` with a
-`PauliLCU` of the same Hamiltonian (alpha, term count, structure), sweeps
-factorization truncation, and measures Chebyshev moments through the
-shared `Walk` consumer.
+runs the encoding on real molecules (integrals from PySCF —
+`pip install pyscf`), across a menu of configurations:
+
+```
+python3 df_block_encoding.py [config]
+
+h2         H2 / STO-3G           2 orbitals ->  4 system qubits  (default)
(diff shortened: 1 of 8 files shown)


ivanbasov (commented):
The move is clean: the API removal is disclosed, the CI paths-filter now covers the example, the pin tests hold the inlined copies to their contract, and the new demo is a big step up from the toy system it replaces. Four inline comments; the two on the demo's self-checks are worth fixing before merge — both can hard-exit a correct run on user-supplied input.

Two small notes on the description:
- "Wheel contents are unchanged" isn't quite right: `df_encoding.py` lived inside the package, so the wheel loses that module — the disclosed API removal, but the artifact does change.
- `compressed_double_factorization` already has example coverage on main (`examples/double_factorization/double_factorization.py`); what's new here is the first *regularized* (RC-DF) use.

examples/bring_your_own_encoding/df_block_encoding.py
```diff
@@ -0,0 +1,398 @@
...
+    print("\n  Truncating the factorization (the knob PauliLCU lacks):")
+    total = factorization.num_leaves
+    if total <= 12:
+        leaf_counts = list(range(1, total + 1))
+    else:
+        leaf_counts = sorted(
+            set(np.linspace(1, total, 10).astype(int).tolist()))
+    errors = []
+    for leaves in leaf_counts:
+        truncated = df.explicit_double_factorization(eri,
```
ivanbasov, line 255: The headline factorization uses `threshold=0.0` (line 234), but the sweep calls here and the RC-DF baselines (lines 288-289) omit `threshold` and get the default `1e-8` — a second, independent stop condition. So sweep rows can be labeled with more leaves than were actually used, and the `errors[-1] < 1e-8` check at line 267 runs on the 1e-8-truncated factorization, where it isn't structurally guaranteed — a correct run on a user-supplied molecule can die in `sys.exit(1)`. Suggest passing `threshold=0.0` in both spots, as `examples/double_factorization/double_factorization.py` already does.

examples/bring_your_own_encoding/df_block_encoding.py
```diff
@@ -0,0 +1,398 @@
...
+                better = f"{ratio:.1f}x better"
+            else:
+                # Near full rank the truncation error is already ~ the
+                # ridge scale, so the regularization bias dominates: the
+                # ridge trades a small fit penalty for bounded cores
+                # (sane alpha). The win to assert is at AGGRESSIVE budgets.
+                better = f"{1.0 / ratio:.1f}x worse (ridge bias; X-DF "\
+                         "already near-exact here)"
+            if xdf_error > 1e-2:
+                wins.append(cdf_error <= xdf_error * 1.001 + 1e-12)
```
ivanbasov, line 315: This asserts an empirical property, not a guaranteed one: L-BFGS monotonically decreases only the *regularized* objective, so the raw fit error can legitimately end more than 0.1% above X-DF whenever the ridge term is non-negligible — the regime the comment at lines 303-309 already describes. The `1e-2` gate is absolute while that regime is scale-invariant, so on a larger molecule or different units (Hartree → eV is ×27.2) a correct run can fail here with `sys.exit(1)`, looking like a library bug. Suggest a relative gate (the sibling example uses error/‖eri‖) or a warning instead of exiting.

tests/python/test_df_encoding.py
```diff
@@ -489,3 +505,33 @@ def test_single_ancilla_negative_sign_select():
...
+    kept = [2e-12, 2e-12, 1e6]
+    total = sum(kept)
+    probabilities = [c / total for c in kept] + [0.0]
+    angles = _prepare_angles(probabilities)
+    assert angles[1] == pytest.approx(2.0 * math.asin(math.sqrt(0.5)))
+    # Exact-zero padding subtrees still produce zero rotations.
+    assert _prepare_angles([0.5, 0.5, 0.0, 0.0])[2] == 0.0
+
+
+def test_walk_kernel_rejects_negative_power():
```
ivanbasov, line 532: The package twin (`test_qubitization.py::test_walk_power_validation`) pins both rejection branches — negative and non-integral — but here only `power=-1` is checked. A copy "simplified" to `int(power)` plus a negativity check would silently truncate fractional powers while this pin stays green. One more line closes the gap:

```python
with pytest.raises(ValueError, match="non-negative integer"):
    encoding.walk_kernel(power=1.5)
```

docs/df_block_encoding.md
```diff
@@ -92,8 +105,28 @@ where the qubits are Z positions *in that frame's rotated basis*).
...
+h2o-cas44  H2O / STO-3G CAS(4,4) 4 orbitals ->  8 system qubits
+h4         linear H4 / STO-3G    4 orbitals ->  8 system qubits
+lih        LiH / STO-3G          6 orbitals -> 12 system qubits
+h2o        H2O / STO-3G          7 orbitals -> 14 system qubits
+h2o-631g   H2O / 6-31G          13 orbitals -> 26 system qubits
+```
+
+Every configuration compares `DoubleFactorizedEncoding` with a `PauliLCU`
+of the same Hamiltonian (alpha, term count, structure), sweeps the
+factorization-truncation dial, and re-optimizes the kept leaves with
```
ivanbasov: `h2o-631g` sets `"cdf": False` and skips the RC-DF rows, so "Every configuration ... re-optimizes the kept leaves with RC-DF" over-claims for one of the six — needs a qualifier like "all but the largest configuration". The module docstring (lines 28-34) repeats the sentence and needs the same tweak.

examples/bring_your_own_encoding/df_block_encoding.py
wsttiger: Good catch — the default `1e-8` threshold was indeed a second stop condition the sweep never intended: rows could be labeled with more leaves than were actually kept, and the full-rank exactness check ran on a factorization where it isn't structurally guaranteed. Both the sweep and the RC-DF baselines now pass `threshold=0.0`, matching the headline factorization (and the sibling example). Fixed in ef96122.

examples/bring_your_own_encoding/df_block_encoding.py
wsttiger: Agreed on both counts: L-BFGS only guarantees descent of the *regularized* objective, so the raw-fit win is empirical, and the absolute `1e-2` gate wasn't scale-invariant. The comparison is now reported rather than asserted — an `OK` line when RC-DF wins everywhere the truncation error is significant, an explanatory note (ridge trades fit for bounded cores; try a smaller regularization) when it trails — and the significance gate is relative to `||eri||_F`. No `sys.exit` on user-supplied molecules. Fixed in ef96122.

tests/python/test_df_encoding.py
wsttiger: Added — the pin now covers both rejection branches (`power=-1` and `power=1.5`), matching the package twin. Fixed in ef96122.

docs/df_block_encoding.md
wsttiger: Right — `h2o-631g` skips it. Both the docs page and the module docstring now say "all but the largest configuration". Fixed in ef96122.

wsttiger:
Thanks for the careful pass — all four inline comments were right, and the two self-check ones were exactly the failure mode worth catching (hard exits on valid user input). All fixed in ef96122; description corrected per your notes on the wheel contents and the existing C-DF example coverage.

ivanbasov (approved):
Verified the fixes in ef96122 and the description corrections — LGTM.

PR #27, opened 2026-08-04
Licensing updates

anjbur:
Update licensing files and headers to match review standards.


wsttiger (commented):
Went through the sweep carefully — the core of it is in great shape. A few comments:

1. **`.github/` YAML consistency**: the sweep converts the `.sh`/`.py` files under `.github/` to the SPDX header, but the 12 YAML files there are split between the old banner (6 workflow files, e.g. `pr_workflow.yaml`) and no header at all (the 3 `action.yaml` files, `copy-pr-bot.yaml`, `cudaq_cache.yaml`, `pr_cache_cleanup.yaml`). Could we apply the same two-line SPDX header to these so there's a single canonical form repo-wide?

2. Nice touch preserving the per-file year ranges (2022–2026 on the CUDA-Q build scripts vs 2026 on newer files) rather than flattening them.

3. On the qsppack NOTICE entry: since upstream publishes no LICENSE file and we're relying on PyPI metadata, do we want an explicit OSRB note/sign-off recorded for that dependency — or should we reach out upstream and ask for a LICENSE file to be added?

4. While we're doing the licensing pass: README.md currently has no license statement at all. Worth adding a short "License: Apache-2.0" section pointing at LICENSE/NOTICE?

5. For the record: I diffed every changed `.py` at the AST level against main and comment-stripped every script/config — no functional changes anywhere — and the full test suite passes on the branch (269 passed, 3 skipped).

anjbur:
I've gone through the requested changes:

1. YAML files are excluded from the OSRB header requirements, which is how these were missed. Great catch. Fully agreed that we should have consistency across them, but I'd actually recommend having no header on any YAML files. That way, we'd match the convention used in cudaqx and cuda-quantum. I've removed the header from all of the `.github` YAML files, but let me know if you would prefer they have the SPDX header. 

3. Both seem like a good idea to me. QSPPACK is neither vendored nor bundled in our wheel and is only separately installed for an optional example/test, so our usage of it is limited and lightweight. Asking if they can add the MIT License file that's referenced on PyPI sounds like a good plan, and in the meantime I'll ask OSRB about this dependency specifically.

4. Great idea. Added.

wsttiger (approved):
LGTM

PR #28, opened 2026-08-11
fix: To fix the displayed version number

kaiqiy-nv:
This PR tries to fix the displayed/returned version of `cudaq-algorithms`.
Since the lib is built CUDA-agnostic in `0.1.0` release, we will not have `cudaq-algorithms-cu12` or `cudaq-algorithms-cu13`, and the user will get `CUDA-Q Algorithms (source)`, which shows no version info and may block some downstream version checks.

### DIFF
--- a/python/cudaq_algorithms/__init__.py
+++ b/python/cudaq_algorithms/__init__.py
@@ -17,7 +17,9 @@ Python package plus NumPy/SciPy.
 def _resolve_version() -> str:
     from importlib.metadata import PackageNotFoundError, version
 
-    for distribution in ("cudaq-algorithms-cu12", "cudaq-algorithms-cu13"):
+    # The shipped wheel is CUDA-agnostic; the -cuNN names are fallbacks.
+    for distribution in ("cudaq-algorithms", "cudaq-algorithms-cu12",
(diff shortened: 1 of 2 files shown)


wsttiger (commented):
Verified the fix end-to-end: a wheel built from base main displays `CUDA-Q Algorithms (source)` after install; the PR wheel displays its real version. The new test genuinely pins the bug — it fails on base and passes on the PR in the installed-wheel lane, while skipping cleanly in the PYTHONPATH/source lane — and the equality assertion (vs. the containment check in the existing `test_import`) is what makes it a real regression test. Nice.

One wording fix before merge: the new comment describes the `cudaq-algorithms-cu12`/`-cu13` names as "future-proofing" — they're the former per-CUDA distribution names kept as fallbacks. Could you reword to just call them fallback names? (Keeping them is fine: the canonical name is tried first, so it wins even if a stale `-cuNN` dist is also present.)

kaiqiy-nv:
Hi @wsttiger , thanks, and updated the comments to say "the -cuNN names are fallbacks".

wsttiger (approved):

PR #29, opened 2026-08-11
fix: Update UCCSD double-excitation handling to correct interleaved open-shell ansatz generation

kaiqiy-nv:
This PR tries to fix [[B] 6592538](https://nvbugspro.nvidia.com/bug/6592538).
It also adds 3 test cases to expose this bug and verify the fix.

In this suggested fix:
We retain the existing CNOT ladder for ordered occupied and virtual ranges, and apply sparse Pauli rotations when their Jordan–Wigner parity intervals overlap. The new path cancels overlapping parity interiors, adjusts the endpoint X/Y operators, and preserves the resulting phase.
In this case, the incorrect direct UCCSD unitary for open-shell systems with `spin >= 2` can be corrected without changing the public API or the physically correct operator pool.

- `test_uccsd_kernel_matches_dense_exponential` (modified):
We add the `(8, 4, 2)` configuration and compare the direct kernel statevector with the dense pool exponential, so errors across the complete interleaved ansatz can be exposed.
- `test_uccsd_interleaved_mixed_double_matches_dense_exponential` (added):
We set only the `[4, 1] -> [3, 6]` mixed-double parameter and compare the kernel result with its dense exponential, so the overlapping-parity failure can be isolated.
- `test_uccsd_spin_two_pool_matches_fermionic_generators` (added):
We compare the complete `(8, 4, 2)` pool with independent dense fermionic generators up to global phase, confirming every spin-2 excitation is present exactly once and physically correct.

### DIFF
--- a/python/cudaq_algorithms/stateprep/_kernels.py
+++ b/python/cudaq_algorithms/stateprep/_kernels.py
@@ -3,12 +3,14 @@
 """State-preparation device kernels (pure Python, composable).
 
 Ports of the ``__qpu__`` kernels in ``lib/stateprep/device/``. The
-``uccsd`` kernel reproduces the C++ excitation enumeration with direct
-index arithmetic (the interleaved-layout orbital indices are affine in the
-loop counters, so no in-kernel lists are needed); gate sequences and the
-parameter order match the C++ exactly. ``uccgsd``, ``upccgsd``, and
-``ceo`` consume the grouped Pauli words and coefficients produced by the
-``get_*_pauli_lists`` helpers, one parameter per group.
+``uccsd`` reproduces the C++ excitation enumeration with direct index
+arithmetic (the interleaved-layout orbital indices are affine in the loop
+counters, so no in-kernel lists are needed). Its parameter order and
+non-interleaved gate sequences match the C++; interleaved doubles use sparse
+Pauli rotations so overlapping Jordan-Wigner strings remain correct.
+``uccgsd``, ``upccgsd``, and ``ceo`` consume the grouped Pauli words and
+coefficients produced by the ``get_*_pauli_lists`` helpers, one parameter per
+group.
 
 ``hartree_fock`` and ``hartree_fock_occupation`` prepare the reference
 determinant those ansatz kernels are applied to, and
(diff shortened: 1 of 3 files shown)


wsttiger (commented):
This is a real and severe bug, and the fix is correct — thank you. I verified `double_excitation` independently against dense ladder-operator exponentials (`expm` of the fermionic generator built from hand-constructed JW matrices) across orderings beyond the ones the tests cover: occupied interval containing both virtuals, virtuals entirely below occupieds, tight interleave, and all four argument-order sign branches — everything agrees to ~1e-15, so the "any ordering of four distinct indices" claim holds. Closed-shell is preserved gate-for-gate (the ordered path is the old circuit unchanged), and the test diff is purely additive.

Two requests before merge:

1. Could you expand the commit message? As written it reads like an enumeration fix, but the enumeration and pools are untouched — the bug is the CNOT parity ladder in `double_excitation` assuming `i_occ < j_occ < a_virt < b_virt`, first reachable at spin >= 2 where the HF occupation interleaves (e.g. `[0, 1, 2, 4]`). Worth also stating that the retired C++ implementation shared the bug — that's why no pinned bit-identical value changes, and reviewers of this module will ask.
2. The spin-2 pool is checked against an independent fermionic generator at only one excitation (`[4,1,3,6]`), while the `(8,4,2)` kernel test uses the pool as its reference. Since `_assert_bijection_up_to_global_phase` already exists in `test_operator_pools.py`, asserting the full spin-2 pool would close that loop cheaply.

Minor: consider squashing — the second commit is a formatting-only touch-up with an identical message.

kaiqiy-nv:
Hi @wsttiger , thanks a lot, and I have made the changes accordingly:
- Expanded the commit message to describe the CNOT parity-ladder assumption, the spin-2 interleaving, the unchanged enumeration/pools, and the retired C++ implementation.
- Replaced the single-excitation check with a full `(8, 4, 2)` pool bijection against independent fermionic generators.
- Squashed the formatting-only commit.
- Updated the PR description in the 'Added/Modified/Deleted test cases' section.

wsttiger (approved):

PR #30, opened 2026-08-11
fix: Use the maximum spin degree for PauliLCU width inference to avoid invalid Pauli-word padding

kaiqiy-nv:
This PR tries to fix [[B] 6592649](https://nvbugspro.nvidia.com/bug/6592649).
It also adds 2 test cases to expose this bug and verify the fix.

In this suggested fix:
We infer the PauliLCU register width from the largest spin-operator degree plus one, and validate an explicit `num_qubits` before rendering Pauli words.
In this case, failures for off-zero or gapped qubit indices can be avoided, while undersized explicit widths produce a clear error.

Added test cases:
- `test_spin_operator_infers_register_extent`:
We use off-zero and gapped spin operators, then check the inferred width and padded Pauli words.
- `test_spin_operator_rejects_explicit_width_below_extent`:
We provide an undersized explicit width, then check that PauliLCU reports the requested and required extents.

### DIFF
--- a/python/cudaq_algorithms/common_kernels.py
+++ b/python/cudaq_algorithms/common_kernels.py
@@ -34,6 +34,29 @@ def state_from(ket) -> cudaq.State:
     return cudaq.State.from_data(np.asarray(ket, dtype=cudaq.complex()))
 
 
+def _maybe_call(value):
+    """Return ``value()`` if callable (property-vs-method API tolerance)."""
+    return value() if callable(value) else value
+
+
+def _term_qubit_extent(term) -> int:
+    """Register extent a spin term needs: largest targeted qubit + 1.
+
+    Not ``qubit_count``: CUDA-Q's ``qubit_count`` is the number of
+    *distinct* targeted indices, which undercounts whenever the targets
+    are off-zero or gapped (``0.5 * spin.x(1)`` targets one qubit but
+    needs a two-qubit word). One helper for every input path (PauliLCU
+    and Trotter both route through it).
+    """
+    try:
+        max_degree = _maybe_call(getattr(term, "max_degree", -1))
+    except RuntimeError:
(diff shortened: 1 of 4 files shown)


wsttiger:
Great catch, and an accurate report — the root cause (`qubit_count` = distinct targets rather than largest index + 1) is exactly right, and the parametrized test pins it nicely. We'd been fixing the same bug in parallel from your QA report and hit one extra wrinkle worth folding in: `term.max_degree` raises `RuntimeError` on scalar identity terms (which the chemistry bridge emits via `scalar_offset`), so the inference as written crashes on bridged Hamiltonians — e.g. `PauliLCU(0.5 * spin.x(1) + 0.25)` (3 `test_df_qsvt_bridge` tests fail). Our first attempt had the identical crash; it's a genuinely subtle API trap.

Since your fork is private I couldn't open a PR into your branch directly, so the delta is one commit on top of your head, pushed here: [`fix/pauli_word_padding_identity_guard`](https://github.com/NVIDIA/cudaq-algorithms/tree/fix/pauli_word_padding_identity_guard) (commit 2a69878 — diff vs. your branch: https://github.com/NVIDIA/cudaq-algorithms/compare/fix/pauli_word_padding_identity_guard~1...fix/pauli_word_padding_identity_guard). It moves the extent inference to a shared guarded helper in `common_kernels.py` (Trotter carried a private copy of the same logic with the same latent crash — it now imports the shared one) and adds three tests on top of yours: a dense-reference check, oversized-width padding, and the identity-term pin. Full suite: 275 passed / 3 skipped with your tests unchanged.

Cherry-pick `2a69878` onto your branch, or — since maintainer edits are enabled — I'm happy to push it straight onto your branch if you prefer. Either way, with that folded in this PR is good to go from my side.

wsttiger (approved):

PR #31, opened 2026-08-12
fix: To include the first-factor eigenvalue scale in second-factor truncation to correct eigendecomposition over-truncation

kaiqiy-nv:
This PR tries to fix [[B] 6595852](https://nvbugspro.nvidia.com/bug/6595852)
It also adds 2 test cases to expose the bug and verify the fix.

In this suggested fix:
We update the second-factor truncation importance calculation to include the absolute first-factor eigenvalue scale before deciding which second-factor eigenmodes to retain.
In this case, significant modes on the first-factor eigendecomposition path can be retained correctly, avoiding scale-independent truncation and large silent reconstruction errors.

- test_second_factor_threshold_includes_first_factor_eigenvalue[numpy]:
We set a rank-one ERI with a first-factor eigenvalue of 100 and a small second-factor mode whose scaled importance exceeds the threshold, and then check that both modes remain and the ERI is reconstructed to round-off.
- test_second_factor_threshold_includes_first_factor_eigenvalue[cupy]:
We run the same scale-sensitive case with the CuPy backend and then check that both modes remain and the ERI is reconstructed to round-off.

### DIFF
--- a/python/cudaq_algorithms/double_factorization/_factorization.py
+++ b/python/cudaq_algorithms/double_factorization/_factorization.py
@@ -144,8 +144,9 @@ def _second_factorization(leaf: DeviceArray, scale: ArrayLike,
     gamma, rotation = xp.linalg.eigh(leaf)
     if second_factor_threshold > 0.0:
         importance = xp.sum(xp.abs(gamma))
-        gamma = xp.where(importance * xp.abs(gamma) > second_factor_threshold,
-                         gamma, 0.0)
+        gamma = xp.where(
+            xp.abs(scale) * importance * xp.abs(gamma)
+            > second_factor_threshold, gamma, 0.0)
     core = scale * xp.outer(gamma, gamma)
     return to_numpy(rotation), to_numpy(core)
 
--- a/tests/python/test_double_factorization.py
+++ b/tests/python/test_double_factorization.py
@@ -495,3 +495,22 @@ def test_pair_swap_asymmetry_rejected():
     eri[0, 0, 1, 1] += 0.05  # breaks (pq|rs) == (rs|pq) only
     with pytest.raises(ValueError, match="chemist symmetries"):
         df.explicit_double_factorization(eri)
+
+
+@pytest.mark.parametrize("backend", _BACKENDS)
+def test_second_factor_threshold_includes_first_factor_eigenvalue(backend):
+    # The small second-factor mode has scaled importance
+    # 100 * (sqrt(0.99) + 0.1) * 0.1 > 0.2, so it must not be truncated.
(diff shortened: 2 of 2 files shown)


wsttiger (commented):
Nice catch — and I like that this makes the two first-factorization paths agree: on the Cholesky path the pivot scale is already baked into the leaf vector's norm, so the threshold was scaled there all along, and only the eigendecomposition path was scale-blind. With `|lambda|` included, the criterion is exactly the mode's contribution to the core one-norm on both paths, which is also what the docstring's "matching OpenFermion's convention" describes (scale absorbed into the one-body squares). I verified your test fails on current main (reconstruction error 14.1 on an ERI of norm 100 with threshold 0.2, dominant mode silently dropped) and that the default Cholesky path is bit-for-bit unchanged.

Full suite is green on your head (273 passed, 4 skipped vs 272/4 on base — the +1 is your new test), the three double-factorization examples still self-verify, and no pinned values move since `second_factor_threshold` defaults to 0.0 and nothing downstream passes it. `xp.abs(scale)` also handles indefinite ERIs correctly, and the change is backend-transparent (scale is a 0-d device array on the CuPy path, a Python float on the Cholesky path — both fine through `xp.abs`).

Two small asks before merge:

1. Could you extend the `explicit_double_factorization` docstring line — "``second_factor_threshold`` optionally zeros small ``gamma^t_k`` (importance-weighted, matching OpenFermion's convention)" — to note that the importance includes ``|lambda_t|`` on the eigendecomposition path? That both-paths-agree semantics is the crux of this fix and worth writing down for the next reader.
2. Optional: in the new test, also asserting that the *unscaled* importance falls below the threshold (e.g. `assert (0.1 + np.sqrt(0.99)) * 0.1 < 0.2`) would pin why 0.2 separates old from new behavior, so the test keeps failing loudly if the scale-free criterion ever sneaks back with different constants.

The rank-one construction in the test is nicely deterministic (no eigh-degeneracy pitfalls) — thanks for parametrizing over both backends. The GPU workflows haven't run yet pending vetting; kicking those off below so the `[cupy]` case actually executes before merge.

wsttiger:
Merging as-is to make today's wheel build — the two small asks from the review above (the docstring note about `|lambda_t|` in the eig-path importance, and the test assertion pinning the threshold separation) will land in a named follow-up commit on main right after the merge. I'd normally have pushed them onto this branch, but the allow-edits grant can't reach a private fork. Both were validated on top of your head with the full suite green (279 passed / 4 skipped). Thanks again for the catch — this one closes a genuinely silent numerical failure.

wsttiger (approved):

PR #32, opened 2026-08-12
fix: To validate FCIDUMP header termination to avoid silently returning zero Hamiltonians

kaiqiy-nv:
This PR tries to fix [[B] 6595943](https://nvbugspro.nvidia.com/bug/6595943).
It also adds 1 test case to expose the bug and verify the fix.

In this suggested fix:
We add an end-state check to `from_fcidump` so that a header without `&END`, `$END`, or `/` raises a clear `ValueError` before zero-filled integral tensors are returned.
In this case, integral records can no longer be misclassified as header text and silently converted into a zero Hamiltonian.

- `test_from_fcidump_rejects_unterminated_header` (added):
We remove `&END` from the existing valid H2 FCIDUMP fixture while retaining its integral records, and then check that `from_fcidump` raises a `ValueError` matching `never terminated`, so the silent zero-Hamiltonian bug can be exposed.

### DIFF
--- a/python/cudaq_algorithms/chemistry.py
+++ b/python/cudaq_algorithms/chemistry.py
@@ -283,6 +283,9 @@ def from_fcidump(contents: str) -> tuple[np.ndarray, np.ndarray, float]:
     if state == "pre":
         raise ValueError(
             "FCIDUMP contents must begin with an &FCI header namelist")
+    if state == "header":
+        raise ValueError(
+            "FCIDUMP header namelist was never terminated (&END, $END, or /)")
 
     header = " ".join(header_lines)
     # Unrestricted files store spin-resolved (alpha/beta/mixed) blocks whose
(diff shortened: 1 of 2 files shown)


wsttiger (commented):
Confirmed end to end — your test fails on current main, and running the unterminated fixture directly returns `max|h1| = max|eri| = core = 0.0`: a silent zero Hamiltonian, exactly the failure mode this library's validation rules exist to prevent. The check is in the right place (after the parse loop, before any tensor work), the error message names all three accepted terminators, and the test can't pass vacuously — if the fixture's `&END` weren't matched, parsing would succeed and `pytest.raises` would fail. Full suite is green on your head (279 passed / 4 skipped; the +1 is your test).

While reproducing this I found the sibling case: a *terminated* header with no integral records also returns an all-zero Hamiltonian silently — `from_fcidump("&FCI NORB=2,NELEC=2,MS2=0,\n&END\n")` gives `max|h1| = 0.0, core = 0.0` on both main and this branch, since NORB parses and the record loop never runs. An FCIDUMP with zero records is never meaningful, so it deserves the same loud rejection.

We're merging this as-is to make today's wheel build; the one-line guard for the sibling case (plus its test) will land in a named follow-up commit on main right after — I'd have pushed it onto this branch, but the allow-edits grant can't reach a private fork. It was validated on top of your head with the full suite green (280 passed / 4 skipped). Nice work tracking this one down.

wsttiger (approved):

PR #33, opened 2026-08-12
Add the Sphinx documentation site; relocate examples into the docs tree and run them in CI

wsttiger:
## Description

A complete Read-the-Docs-style documentation site for cudaq-algorithms,
modeled on the cudaqx docs (RTD theme, autodoc API reference, guides,
runnable examples rendered from a single source of truth). The loose
`docs/*.md` pages are migrated into the site and deleted; all runnable
examples move into the docs tree so a documentation page can never drift
from the code it renders — and CI now executes every one of them.

### The site

- **`docs/sphinx/`** with a checked-in `conf.py` (RTD theme, NVIDIA
  green; autodoc/autosummary/napoleon, MyST, MathJax, copybutton,
  inline-tabs, intersphinx) and an 8-section toctree: Getting Started
  (install, quickstart, a 7-example learning series), guides
  (preprocessing, state preparation, block encodings, Trotter,
  qubitization & QSVT), Examples, Conventions, and an autodoc Python API
  reference.
- `make -C docs html` builds under **`-W --keep-going`** (warnings are
  errors) — that gate forced docstring cleanups across the package
  (unescaped ket notation, an unparseable `Parameters` block, two
  undocumented `Trotter` properties), so the API reference renders clean.
- Displayed equations are real math (MathJax): the block-encoding
  definition, the second-quantized Hamiltonian, the Jordan–Wigner ladder
  operators, the double-factorization form, Trotter error orders. Every
  formula was validated through the MathJax v3 parser, not just the
  Sphinx build (MathJax renders client-side, so `-W` alone cannot catch
  LaTeX errors).
- House rules applied throughout: guides document *usage* —
  implementation and test-suite narration stays with the code.

### Examples: one home, CI-enforced

- All runnable examples live in **`docs/sphinx/examples/python/`**
  (cudaqx-style), `literalinclude`d into the docs via
  `[Begin Documentation]` markers. The repo-root `examples/` directory
  is gone.
- The bring-your-own-encoding pair from #26 (`df_encoding.py` + the
  PySCF molecule-menu demo) relocates with everything else;
  `tests/python/test_df_encoding.py` and the `pr_workflow.yaml` paths
  filters follow (`docs/sphinx/examples/python/**` in both
  `build-algorithms` and `build-wheels`) — reviewers: this is the one
  place the branch touches CI filter logic, and it is the direct
  continuation of the filter #26 added.
- **New: every documentation example runs in CI.**
  `tests/python/test_examples.py` executes each script as a subprocess
  and requires its self-checks to pass (~44 s for all 15; per-example
  timeout; a discovery guard fails if the glob ever finds nothing).
  `qsppack` joins `pyscf` in the CI dependency installs. An
  example-only PR now executes every example.
- **New example 7 — a QSVT linear solver**: invert a 5×5 SPD system via
  a Childs–Kothari–Somma polynomial approximating 1/x (the QSVT reading
  of HHL), same `PauliLCU` + `QSVT` machinery as example 2's time
  evolution with only the phase sequence changed. Self-verified against
  `numpy.linalg.solve` (fidelity 0.999999, residual ~1e-3).
- The Givens-rotation Slater-determinant preparation now cites its
  sources (Jiang et al., PRApplied 9, 044036 (2018); Kivlichan et al.,
  PRL 120, 110501 (2018)) in the module docstring and the guide.

### Docs CI

`.github/workflows/docs.yaml`: pure-Python build (pip cudaq +
`docs/requirements.txt`, `make html` under `-W`), Pages artifact upload,
deploy on `releases/v*`. Note this workflow has not yet executed on
GitHub (it triggers on main-push/PR/dispatch); a `workflow_dispatch` run
after merge is the first live test.

### Compatibility notes

- Wheel contents unchanged — docs, tests, and examples were never
  packaged.
- `README.md` updated to point at the site and the new examples path.
- The Getting Started install section notes that
  `pip install cudaq-algorithms` applies once the wheel is published,
  with the find-links/source alternatives until then.
- Coordination item (outside this branch): `CLAUDE.md` (tracked on the
  getting-started branch) still describes the old layout (`examples/`,
  `docs/*.md`); it needs an update when either branch lands.


docs/sphinx/examples_rst/block_encodings.rst
```diff
@@ -0,0 +1,43 @@
+Block-encoding examples
```
melody-ren, line 1: In general, it'd be nice to have each example links back to it corresponding User Guide section, if there's one. Something like a .. sealso:: block or a embedded link pointing back to the User Guide's page. Likewise, each section in User guide page should also link to the example

docs/sphinx/guide/qubitization_qsvt.rst
```diff
@@ -0,0 +1,181 @@
+Qubitization and QSVT
```
melody-ren, line 1: Maybe I just missed it. PR description says that `07_matrix_inversion_qsvt.py` is a major addition but this guide doesn't mention it here

docs/sphinx/index.rst
```diff
@@ -0,0 +1,42 @@
...
+kernels.
+
+.. toctree::
+   :caption: Getting Started
+   :maxdepth: 2
+
+   getting_started/index
+
+.. toctree::
+   :caption: User Guide
```
melody-ren, line 18: The guide pages reference examples by raw file-path strings (backtick-quoted paths) rather than Sphinx `:doc:` cross-references, so they render as inert code text instead of clickable links. `guide/block_encodings.rst` already does this correctly on line 189 with `:doc:`../examples_rst/block_encodings`` — the other pages should match that pattern:

| File | Line | Current | Should be |
|---|---|---|---|
| `guide/state_prep.rst` | 36 | raw path to `hartree_fock_ucc.py` | `:doc:`../examples_rst/state_prep`` |
| `guide/trotter.rst` | 23 | raw path to `trotter_chemistry.py` | `:doc:`../examples_rst/hamiltonian_simulation`` |
| `guide/qubitization_qsvt.rst` | 116 | raw path to `hamiltonian_simulation_qsvt.py` | `:doc:`../examples_rst/hamiltonian_simulation`` |
| `guide/preprocessing.rst` | 180 | raw ref to `03_chemistry_to_ground_state.py` | `:doc:`../examples_rst/getting_started`` |
| `guide/preprocessing.rst` | 378 | raw paths to `double_factorization.py` etc. | `:doc:`../examples_rst/preprocessing`` |
| `guide/block_encodings.rst` | 154 | raw path to `06_bring_your_own_encoding.py` | `:doc:`../examples_rst/getting_started`` |
| `guide/block_encodings.rst` | 166 | raw path to `df_encoding.py` | `:doc:`../examples_rst/block_encodings`` |

melody-ren (commented):
Thanks Scott! Just some notes about linking pages

docs/sphinx/getting_started/index.rst
```diff
@@ -0,0 +1,73 @@
+***************
+Getting Started
+***************
+
+CUDA-Q Algorithms is a **primitive-first** library of fault-tolerant quantum
+algorithms built on `CUDA-Q <https://nvidia.github.io/cuda-quantum/>`_. It is
+pure Python: quantum primitives are ``@cudaq.kernel`` device functions, and the
+classical preprocessing is NumPy/SciPy.
+
```
eliotheinrich, line 9: I think the getting started page here would benefit from an introduction to the main components of the library, how they fit together, and what the ultimate goals of the repo are, in plain English. This would be helpful grounding for users without pre-existing context about the repo.

docs/sphinx/getting_started/index.rst
```diff
@@ -0,0 +1,73 @@
...
+   $ PYTHONPATH=python python3 -m pytest tests/python -q
+
+The dense-reference tolerances require an fp64 simulator; ``conftest.py``
+selects ``qpp-cpu`` by default (override with ``CUDAQ_DEFAULT_SIMULATOR``).
+
+The primitive-first philosophy
+==============================
+
+Every algorithm is exposed as a **factory that emits a CUDA-Q kernel**: the
+constructor holds the problem, and a method call holds the choices.
```
eliotheinrich: Maybe link to CUDA-Q docs for users unfamiliar with the parlance?

docs/sphinx/guide/state_prep.rst
```diff
@@ -0,0 +1,287 @@
...
+     __init__.py       the public surface (everything re-exported here)
+     _kernels.py       device kernels: uccsd, uccgsd, upccgsd, ceo,
+                       single/double excitations, hartree_fock,
+                       hartree_fock_occupation, fixed_parameter_ucc
+     _pools.py         host side: excitation enumeration, operator pools,
+                       grouped Pauli-list converters
+     _hartree_fock.py  host side: occupation builders, pool-to-Pauli-list
+                       conversion, hartree_fock_ucc_kernel factory,
+                       resource estimators
+     _givens.py        Givens-rotation Slater determinants
```
eliotheinrich: Should _-prefixed files be advertised here? The internal file structure should be irrelevant and opaque for users who only consume __init__.py

docs/sphinx/guide/state_prep.rst
```diff
@@ -0,0 +1,287 @@
...
+`(qubits: cudaq.qview)` kernel, which makes it directly injectable into the
+algorithm factories via their `state_prep=` parameter — the block-encoding
+factories (:doc:`block_encodings`), the Trotter primitive (:doc:`trotter`),
+and qubitization / QSVT (:doc:`qubitization_qsvt`) all accept it. See
+`State preparation injection`_ below for the shared contract.
+
+Chemistry-style state-preparation kernels and their host-side helpers are
+implemented as a pure-Python peer of the LCU/QSVT and Trotter primitives:
+UCC-family ansatz kernels, operator pools and Pauli-list converters,
+Hartree-Fock reference determinants, fixed-parameter UCC products, and
```
eliotheinrich: UCC is undefined here

docs/sphinx/guide/state_prep.rst
```diff
@@ -0,0 +1,287 @@
+State preparation
+=================
+
+State preparation is an **input** to the library's algorithms, not an
+algorithm in itself: every preparation here is (or can be packaged as) a
+`(qubits: cudaq.qview)` kernel, which makes it directly injectable into the
```
eliotheinrich: I think this means a kernel that takes `qubits: cuda.qview` as an argument, but this could perhaps be made explicit here.

docs/sphinx/examples/python/05_state_prep_and_injection.py
```diff
@@ -0,0 +1,99 @@
...
+# This source code and the accompanying materials are made available under     #
+# the terms of the Apache License 2.0 which accompanies this distribution.     #
+# ============================================================================ #
+# [Begin Documentation]
+"""Example 5 — State preparation and injection.
+
+Every primitive factory (Walk, QSVT, Trotter) takes a `state_prep`
+argument: a `(qubits: qview)` CUDA-Q kernel that prepares the register.
+Pass one and the factory returns a *zero-argument* circuit -- no
+`cudaq.State`, no statevector anywhere -- that you can sample directly
```
eliotheinrich: The 'no statevector anywhere' phrase is repeated in a few places, but I think it's more accurate to say that the statevector is abstracted from the caller; correct me if I'm wrong, but a classical backend will still need a statevector, and this statement could give the false impression that no statevector is involved anywhere in the stack.

docs/sphinx/guide/state_prep.rst
```diff
@@ -0,0 +1,287 @@
...
+  and
+  odd-electrons-at-spin-0 in `get_uccsd_excitations`); every guard added
+  in the pure implementation raises `ValueError`.
+- **CEO orbital counts**: the CEO helpers (`make_ceo_operator_pool`,
+  `get_ceo_pauli_lists`) take `num_orbitals` in *spatial* orbitals — the
+  pool acts on `2 * num_orbitals` qubits.
+  Everything else in the package counts qubits (spin orbitals).
+
+See :doc:`../conventions` for the library-wide qubit-ordering, Pauli-word,
+and integral-tensor conventions these build on.
```
eliotheinrich, line 271: Should this be moved to the dedicated `conventions.rst`?

docs/sphinx/guide/state_prep.rst
```diff
@@ -0,0 +1,287 @@
...
+`docs/sphinx/examples/python/hartree_fock_ucc.py` (fixed-parameter UCCSD vs.
+dense matrix exponentials) and
+`docs/sphinx/examples/python/givens_slater_determinant.py` (Slater
+determinants).
+
+Ansatz kernels and operator pools
+---------------------------------
+
+The `uccsd`, `uccgsd`, `upccgsd`, and `ceo` device kernels are
+`@cudaq.kernel` functions, composable from user kernels, composable from user kernels:
```
eliotheinrich: These acronyms are undefined

docs/sphinx/guide/trotter.rst
```diff
@@ -0,0 +1,148 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+- `order=1` — first-order Trotter: one :math:`\exp(-i c_i (t/\mathrm{steps}) P_i)` sweep per
+  step; error :math:`O(t^2/\mathrm{steps})`.
+- `order=2` (default) — symmetric second-order (Strang) splitting: a
+  half-angle forward sweep followed by a half-angle reverse sweep; error
+  :math:`O(t^3/\mathrm{steps}^2)`.
+- `order=4` — Forest-Ruth fourth-order formula: three symmetric
+  second-order sub-steps with time fractions `w1`, `w0`, `w1`; error
+  :math:`O(t^5/\mathrm{steps}^4)`.
```
eliotheinrich, line 63: There is a mix of inline LaTeX and code blocks expressing mathematical expressions; nitpick, but it is a little jarring. For example, the similar :math:`\exp(-i c_i (t\mathrm{steps}) P_i)` expressions appear later in this page as inlined code blocks.

docs/sphinx/getting_started/index.rst
```diff
@@ -0,0 +1,73 @@
...
+================
+
+- :doc:`../guide/preprocessing` — molecule / integrals to a qubit Hamiltonian.
+- :doc:`../guide/state_prep` — Hartree-Fock, UCC, and Givens state preparation.
+- :doc:`../guide/block_encodings` — ``PauliLCU``, the ``BlockEncoding``
+  protocol, and a worked double-factorized example encoding.
+- :doc:`../guide/trotter` and :doc:`../guide/qubitization_qsvt` — the two
+  Hamiltonian-simulation routes.
+- :doc:`../conventions` — qubit ordering, integral tensors, and normalization
+  conventions (read this before validating any numerics).
```
eliotheinrich: IMO, it seems like `state_prep` should maybe go lower in the order here (and in the toctree ordering); the core library utilities (QSVT, Trotter, etc) should be highlighted first

docs/sphinx/guide/state_prep.rst
eliotheinrich: 'composable from user kernels' is duplicated

eliotheinrich (commented):
Thanks, Scott! Most of my comments are nits or opinions about readability. The only one I feel strongly about is a more substantial introduction for new users.

docs/sphinx/examples_rst/block_encodings.rst
wsttiger: Done in 074359e — guide pages now use `:doc:` links to their examples pages (all seven raw-path references converted), and each examples page opens with a `seealso` back to its guide.

docs/sphinx/guide/qubitization_qsvt.rst
wsttiger: Good catch — the guide now covers `07_matrix_inversion_qsvt.py` right after the time-evolution example, framed as the same machinery with only the phase sequence changed.

docs/sphinx/index.rst
wsttiger: Applied exactly as tabulated — thanks for the precise list.

docs/sphinx/getting_started/index.rst
wsttiger: Agreed this was missing. Getting Started now opens with a "What's in the library" section: the four layers (chemistry preprocessing → block encodings → algorithm primitives → state preparation), how they feed each other, and the goal of the repo in plain English, ending with the molecule-to-spectra path through the library.

docs/sphinx/getting_started/index.rst
wsttiger: Linked CUDA-Q's kernel-basics page at the first use of "CUDA-Q kernel".

docs/sphinx/guide/state_prep.rst
wsttiger: You're right — internal layout removed; the guide now describes only the `stateprep` namespace users import.

docs/sphinx/guide/state_prep.rst
wsttiger: Expanded at first use — "unitary-coupled-cluster (UCC)".

docs/sphinx/guide/state_prep.rst
wsttiger: Made explicit: "a CUDA-Q kernel taking the qubit register as its only argument (`qubits: cudaq.qview`)".

docs/sphinx/examples/python/05_state_prep_and_injection.py
wsttiger: You're correct — a classical backend still holds a statevector internally, so "anywhere" overclaimed. Rephrased at all five occurrences: the statevector is abstracted from the caller / never crosses the API boundary.

docs/sphinx/guide/state_prep.rst
wsttiger: Kept it here deliberately: `conventions.rst` holds the library-wide conventions (qubit ordering, Pauli words, integral tensors), while this one is scoped to the CEO helpers alone — they're the only APIs in the package that count spatial orbitals rather than qubits, so the warning is most useful at the point of use. The section's closing paragraph already cross-links to `conventions.rst` for everything library-wide.

docs/sphinx/guide/state_prep.rst
wsttiger: All expanded at first use: UCCSD (unitary coupled cluster with singles and doubles), UCCGSD (generalized), UpCCGSD (paired generalized doubles), CEO (coupled exchange operator).

docs/sphinx/guide/trotter.rst
wsttiger: Agreed — converted the remaining code-styled math in this guide (the Forest-Ruth weights and the `exp(-ict)` phase notes) to LaTeX.

docs/sphinx/getting_started/index.rst
wsttiger: Reordered both the "Where to go next" list and the guide toctree: block encodings and the two simulation primitives first, state prep after.

docs/sphinx/guide/state_prep.rst
wsttiger: Fixed.

wsttiger:
Thanks both — all feedback addressed in 074359e (replies threaded on each comment). Docs still build under `-W` and the full suite is green.

/ok to test 074359e9f23f0be749fb747bc40b795a8b278ac5

melody-ren (approved):

PR #37, opened 2026-08-17
Accept the resolver's CUDA variant on CPU wheel-test lanes

wsttiger:
The v0.1.0 release wheel run failed in both CPU test lanes: PyPI mode lets the released `cudaq` metapackage resolve the `cuda-quantum-cuNN` variant, and cudaq 0.15.1 changed the no-driver default to the newest CUDA (cu13), breaking the CPU lanes' pinned `cu12` expectation (the `cuda_toolkit[cudart]` nudge is ignored by the new resolver — confirmed in the run logs, which show the cudart install succeeding and cu13 resolving anyway).

Fix: the exact-variant check stays where it is meaningful — Custom mode (we installed the variant explicitly) and GPU lanes (detection must match the host). CPU lanes in PyPI mode now assert exactly one `cuda-quantum-cuNN` resolved, whichever it is: the resolver's no-driver policy belongs to cudaq, not to this wheel's contract, and the wheel is CUDA-agnostic by design. The lane's `device` rides into the script as a new argument; the inert cudart pre-install is removed.

Verified: branch logic exercised for all three lane shapes (cpu-pypi → any-variant, gpu-pypi → exact, custom → exact); `bash -n` clean; single invocation site updated.

After merge: fast-forward `releases/v0.1.0` to main and re-dispatch the wheels workflow.

### DIFF
--- a/.github/workflows/wheel_algorithms.yaml
+++ b/.github/workflows/wheel_algorithms.yaml
@@ -251,4 +251,5 @@ jobs:
             ${{ matrix.platform }} \
             ${{ matrix.cuda_version }} \
             ${{ inputs.cudaq_wheels == 'Custom' && '0.15.99' || 'SKIP' }} \
-            ${{ inputs.version || '0.15.99' }}
+            ${{ inputs.version || '0.15.99' }} \
+            ${{ matrix.device }}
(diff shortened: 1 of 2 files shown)


anjbur (approved):
LGTM!

PR #38, opened 2026-08-18
fix: To record C-DF optimizer status and cache CG warm starts by parameter vector to avoid silent non-convergence and inconsistent L-BFGS gradients

kaiqiy-nv:
This PR tries to fix [[B] 6627715](https://nvbugspro.nvidia.com/bug/6627715).
It also adds 4 test cases to expose this bug and verify the fix.
The examples and docs are also updated accordingly.

(PR open for maintainer edits. Hope this time it can work.)
(Tests and examples passed on local machine.)

In this suggested fix:
We record L-BFGS-B `success` / `nit` / gradient norm on the returned `DoubleFactorization`, warn when the run does not converge, and key the CG warm start by the parameter vector so a rejected line-search evaluation cannot change `(f, g)` at the same `x`.
In this case, silently returning an under-converged C-DF (indistinguishable from a stationary point) and a history-dependent CG objective (which breaks the L-BFGS-B contract) can be avoided.

- `test_cdf_reports_when_optimizer_hits_iteration_limit`:
We set a truncated synthetic ERI with `max_iterations=1` and then check that a `RuntimeWarning` matching `did not converge` is raised and that `optimizer_success` is False with `optimizer_nit == 1`, so a cap-hit can no longer look like success.
- `test_cdf_exact_rank_is_marked_converged`:
We set a synthetic ERI at its true leaf rank (X-DF warm start already minimises) and then check that no `RuntimeWarning` is issued and `optimizer_success` is True with `optimizer_nit == 0`, so a genuine stationary point is still reported as converged.
- `test_xdf_leaves_optimizer_status_unset`:
We run X-DF (no L-BFGS) and then check that `optimizer_success`, `optimizer_nit`, and `optimizer_grad_norm` remain `None`, so the new fields do not leak onto the explicit path.
- `test_cdf_objective_is_a_function_of_x`:
We wrap the real C-DF objective, evaluate at `x`, at a rejected trial `x+dx`, then at `x` again (H2O/STO-3G, 4 leaves, both `lstsq` and `cg`), and then compare `|f(x)-f(x)_again|` and `||g-g_again||` to tight tolerances, so a CG warm start that reused the trial cores would be exposed.

Modified examples:
- `docs/sphinx/examples/python/double_factorization.py`:
We raise the H2O C-DF L-BFGS budget from 600 to 5000 (truncated molecular C-DF can need a few thousand steps) and print `(not converged)` from `optimizer_success`, so the advertised table no longer reports a silently truncated iterate.
- `docs/sphinx/examples/python/df_block_encoding.py`:
We drop the explicit `max_iterations=300` so RC-DF uses the library default (2000) and will warn if L-BFGS still hits the cap.

Modified docs:
- `docs/sphinx/guide/preprocessing.rst`:
We document the C-DF `optimizer_*` fields, the non-convergence `RuntimeWarning`, and that CG warm start reuses cached cores on re-evaluation of the same parameters, so callers know to inspect status rather than treat every returned factorization as a stationary point.

Logs from running modified examples:
```
===== double_factorization.py =====
Double factorization of H2O/STO-3G two-electron integrals
  backend: cupy
  orbitals: 7   ||eri||_F: 7.229640

X-DF (explicit):
  full-rank leaves: 28   reconstruction error: 6.951e-15
  threshold 1e-02: leaves=18  rel error=4.235e-03
  threshold 1e-03: leaves=23  rel error=2.285e-04
  threshold 1e-04: leaves=25  rel error=6.079e-06

C-DF (compressed) vs X-DF at equal leaf count:
  leaves= 2:  X-DF rel error=1.686e-01   C-DF rel error=5.958e-02
  leaves= 4:  X-DF rel error=1.228e-01   C-DF rel error=2.172e-02
EXIT_DF:0
===== df_block_encoding.py (default h2) =====
=== h2: H2 / STO-3G ===
RHF energy -1.116684 Ha; core/nuclear constant 0.713754 Ha (added classically, not encoded)
2 spatial orbitals -> 4 system qubits

  flat PauliLCU:            alpha =     2.6977, 15 Pauli terms, 4 ancillas
  DoubleFactorizedEncoding: alpha =     2.8790, 18 Z-word terms in 4 frames (3 leaves), 4 Givens rotations, 5 ancillas

  Truncating the factorization (the knob PauliLCU lacks):
      1 leaves: alpha =     2.5380,    11 terms, tensor error 3.65e-01
      2 leaves: alpha =     2.9006,    17 terms, tensor error 4.33e-02
      3 leaves: alpha =     2.8790,    18 terms, tensor error 3.42e-16
  [check] truncation error is non-increasing in leaves ... OK
  [check] full-rank X-DF reconstructs the ERI exactly ... OK

  RC-DF at the same leaf budgets (optimize the kept leaves, don't just truncate):
      1 leaves: X-DF error 3.65e-01  RC-DF error 3.63e-01  (1.0x better), RC-DF alpha = 2.5164
      2 leaves: X-DF error 4.33e-02  RC-DF error 1.53e-04  (282.5x better), RC-DF alpha = 2.5164
  [check] RC-DF fits at least as well wherever truncation error is still significant ... OK

  Circuits: 4 system + 5 ancilla = 9 qubits
  encoded block vs sparse JW reference: max |diff| = 6.09e-16
  [check] <0|U|0> == H/alpha ... OK

  Even Chebyshev moments <T_k(H/alpha)> through Walk (same
  consumer, either encoding; alphas differ, H is the same):
    T_0:  PauliLCU +1.00000000 (exact +1.00000000)   DF +1.00000000 (exact +1.00000000)
  [check] T_0 moments match the reference ... OK
    T_2:  PauliLCU -0.67262461 (exact -0.67262461)   DF -0.71255633 (exact -0.71255633)
  [check] T_2 moments match the reference ... OK
    T_4:  PauliLCU +0.03888226 (exact +0.03888226)   DF +0.11880394 (exact +0.11880394)
  [check] T_4 moments match the reference ... OK
EXIT_BE:0
```

### DIFF
--- a/docs/sphinx/examples/python/df_block_encoding.py
+++ b/docs/sphinx/examples/python/df_block_encoding.py
@@ -289,10 +289,7 @@ def run(key: str, force_circuits: bool):
                                                          max_num_leaves=leaves)
             xdf_error = df.factorization_error(eri, truncated)
             compressed = df.compressed_double_factorization(
-                eri,
-                num_leaves=leaves,
-                regularization=1e-4,
-                max_iterations=300)
+                eri, num_leaves=leaves, regularization=1e-4)
             cdf_error = df.factorization_error(eri, compressed)
             cdf_alpha = DoubleFactorizedEncoding(one_body, compressed).alpha
             ratio = xdf_error / max(cdf_error, 1e-16)
--- a/docs/sphinx/examples/python/double_factorization.py
+++ b/docs/sphinx/examples/python/double_factorization.py
@@ -64,14 +64,18 @@ def main():
                                                     threshold=0.0,
                                                     max_num_leaves=num_leaves,
                                                     backend=backend)
+        # Truncated molecular C-DF can need a few thousand L-BFGS steps;
+        # the library default (2000) sometimes hits the cap. Status is
+        # still reported if this budget is not enough.
         compressed = df.compressed_double_factorization(eri,
                                                         num_leaves=num_leaves,
-                                                        max_iterations=600,
+                                                        max_iterations=5000,
                                                         backend=backend)
         x_rel = df.factorization_error(eri, explicit) / norm
         c_rel = df.factorization_error(eri, compressed) / norm
+        status = "" if compressed.optimizer_success else "  (not converged)"
         print(f"  leaves={num_leaves:2d}:  X-DF rel error={x_rel:.3e}   "
-              f"C-DF rel error={c_rel:.3e}")
+              f"C-DF rel error={c_rel:.3e}{status}")
         assert c_rel <= x_rel + 1.0e-6
 
 
--- a/docs/sphinx/guide/preprocessing.rst
+++ b/docs/sphinx/guide/preprocessing.rst
@@ -310,8 +310,8 @@ accuracy:
 - **`cg_warm_start=True`** seeds each *new* L-BFGS evaluation's CG from the
   previous evaluation's cores. The cores move slowly between steps, so the
   warm start collapses the iteration count. A re-evaluation of the same
-  parameters (a rejected line-search point) reuses the cached cores, so
-  the objective stays a function of the parameters.
+  parameters (a rejected line-search point) reuses cores from a short LRU
+  of recent solves, so the objective stays a function of the parameters.
 - **`cg_optimization_tolerance`** (default ``max(cg_tolerance, 1e-6)``)
   solves the *in-loop* systems only loosely — an inexact inner solve, which
   is sound here because the envelope theorem only needs an approximate
--- a/docs/sphinx/guide/preprocessing.rst
+++ b/docs/sphinx/guide/preprocessing.rst
@@ -255,7 +255,10 @@ X-DF vs C-DF
   ``X^t``) and optimized with L-BFGS, while the symmetric cores are solved
   in closed form at each step (the two-step scheme of the paper,
   warm-started from X-DF). C-DF reaches a target accuracy with substantially
-  fewer leaves than X-DF.
+  fewer leaves than X-DF. The L-BFGS status is stored on the result
+  (``optimizer_success``, ``optimizer_nit``, ``optimizer_grad_norm``); a
+  ``RuntimeWarning`` is issued if the run hits ``max_iterations`` or otherwise
+  fails to converge, and the best factorization found is still returned.
 
 RC-DF regularization
 ~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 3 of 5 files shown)


wsttiger (commented):
Nice diagnosis — the x1 -> x2 -> x1 replay test is exactly the right probe, and I confirmed it catches a real 1.8e-5 objective inconsistency on main that your fix removes without perturbing any real trajectory (bit-identical base-vs-PR leaf cores under pinned BLAS threads on 6 seeded lstsq/cg cases). The status recording matches the module's existing warn-for-quality / raise-for-invalid-input split (the indefinite-ERI warning is the precedent), the `preprocessing.rst` updates are accurate against the code, and the example surfacing `(not converged)` closes the loop nicely. Full suite is 317 passed / 3 skipped on your head; yapf and DCO clean.

One request before merge, and two small nits:

1. **The `by_x` cache grows without bound.** Each distinct x evaluated adds a full cores-stack: on an H2O/STO-3G cg run I count 1,398 entries and zero cache hits (SciPy's `MemoizeJac` already absorbs the immediate same-x repeat, so hits deeper than that are rare in practice). At production sizes that's multi-GB — and on the cupy backend these are *device* arrays, so it's GPU memory. Keeping just the most recent `(key, cores)` pair — or a 2–3 entry LRU — gives the identical same-x guarantee your test pins down, with O(1) memory.

2. `test_cdf_exact_rank_is_marked_converged` asserts `optimizer_nit == 0`, which encodes L-BFGS-B internals that have shifted across scipy releases (scipy is unpinned in pyproject). `optimizer_success is True` plus a small `optimizer_grad_norm` is the same intent, less version-coupled.

3. In `test_cdf_objective_is_a_function_of_x`, the replay-through-`minimize` trick is a genuinely good pattern — one suggestion: assigning the scipy module attribute patches it process-globally (the `finally` covers it today); `monkeypatch.setattr` gives the same effect with guaranteed scoped restoration.

Side observation from verifying this PR, not your problem to fix: with BLAS threads unpinned, C-DF is nondeterministic run-to-run on *identical* code (reduction-order noise compounds over ~2000 L-BFGS steps into different local minima — I measured 6.7e-2 vs 8.0e-2 reconstruction error on the same base commit). Anyone A/B-testing C-DF changes should pin `OMP/OPENBLAS_NUM_THREADS=1` first; we'll add a note to the conventions page.

wsttiger (approved):

kaiqiy-nv:
Hi @wsttiger , I have updated the code accordingly.

For [1], we implemented a 3-slot LRU of `(x, cores)` instead of the unbounded `by_x` dict (one slot is not enough for `x → x+Δ → x`). Hits reuse cores; misses still warm-start from the last `z`. Docs updated to match.

For [2], we dropped `optimizer_nit == 0` and assert `optimizer_grad_norm < 1e-8`, keeping success / no-warning / reconstruction error.

For [3], the replay test now uses `monkeypatch.setattr`.

The unpinned BLAS threads issue is left for future PRs.

PR #40, opened 2026-08-25
Add unary-iteration and QROM primitives

wsttiger:
## Summary

Adds the first primitives to `cudaq_algorithms.primitives`:

- Unary iteration for applying an address-dependent operation
- QROM for coherent table lookup

These primitives provide the foundation for the THC block encoding and future SELECT-style constructions.

All implementations in this PR are strictly unitary. They use no mid-circuit measurement or classical feedback, compose under `cudaq.control`, and support CUDA-Q sampling, statevector, and observable workflows.

## Unary iteration

Adds `unary_iteration_kernels`, a factory for constructing address-controlled operations from a callback that emits the operation for each address.

The default fused walker updates the unary ladder in place instead of uncomputing and recomputing it. For a complete sweep over `N` addresses, this reduces the Toffoli count from `2N - 4` for the conventional compute/uncompute construction to `3N/2 - 5`.

The controlled construction uses `3N/2 - 1` Toffolis.

This is the strictly unitary select(V) walk of Childs et al. (arXiv:1711.10980, Appendix G.4, Lemma G.7); the lower `N - 1` unary-iteration count of Babbush et al. (arXiv:1805.03662) requires measurement-based uncomputation (Gidney, arXiv:1709.06648), which this PR deliberately excludes.

The factory also provides:

- Explicit inverse kernels
- `cudaq.control` compatibility
- `describe()` output for inspecting generated operations
- Analytical `toffoli_count` bookkeeping verified against compiled circuits

## QROM

Adds:

```python
QROM(data, address_bits, output_bits)
```

with two lookup strategies:

- `select`: lookup using unary iteration
- `select_swap`: SELECT-SWAP/QROAM using clean block ancillas

With `variant="auto"`, QROM evaluates the available strategies and selects the lowest-cost variant and block size for the supplied table dimensions.

### Example resource costs

For a 10-bit address register and `N = 1024`:

| Construction | Toffolis |
|---|---:|
| Conventional unary walker | 2044 |
| Fused unary walker | 1531 |
| QROM `select` | 1531 |
| QROM `select_swap`, 8-bit output | 422 |
| QROM `select_swap`, 16-bit output | 598 |

All reported counts are checked against the compiled circuits using `cudaq.estimate_resources`.

## Validation

The tests cover:

- Exhaustive unary-iteration truth tables through six address bits
- All supported unary tree shapes
- Dense statevector references
- Random QROM tables
- Power-of-two and non-power-of-two table sizes
- Both QROM strategies
- Automatic strategy and block-size selection around crossover points
- Multi-qubit, noncommuting SELECT payloads
- Phased-superposition address states
- Forward and inverse circuit behavior
- Equality between analytical and compiler-reported Toffoli counts

Test result at the current branch tip:

```text
391 passed, 3 skipped
```

## References

- Babbush et al., [arXiv:1805.03662](https://arxiv.org/abs/1805.03662)
- Gidney, [arXiv:1709.06648](https://arxiv.org/abs/1709.06648)
- Low, Kliuchnikov, and Schaeffer, [arXiv:1812.00954](https://arxiv.org/abs/1812.00954)

### DIFF
--- a/python/cudaq_algorithms/primitives/__init__.py
+++ b/python/cudaq_algorithms/primitives/__init__.py
@@ -2,9 +2,12 @@
 # SPDX-License-Identifier: Apache-2.0
 """Fault-tolerant circuit primitives: unary iteration and QROM.
 
-``unary_iteration_kernels`` mints the fused tree walk of Babbush et al.
-(`arXiv:1805.03662`, Fig. 7) — apply a per-address body exactly when the
-address register equals ``k`` — and ``QROM`` builds coherent
+``unary_iteration_kernels`` mints unary iteration (Babbush et al.,
+`arXiv:1805.03662`, Fig. 7) in its strictly unitary form — no
+measurement-based uncomputation, i.e. the fused select(V) walk of
+Childs et al. (`arXiv:1711.10980`, Appendix G.4) — applying a
+per-address body exactly when the address register equals ``k``.
+``QROM`` builds coherent
 classical-table lookups on top of it, either as the plain
 unary-iteration SELECT or as the SELECT-SWAP / QROAM construction of
 Low, Kliuchnikov and Schaeffer (`arXiv:1812.00954`).
--- a/python/cudaq_algorithms/primitives/__init__.py
+++ b/python/cudaq_algorithms/primitives/__init__.py
@@ -1,0 +1,30 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""Fault-tolerant circuit primitives: unary iteration and QROM.
+
+``unary_iteration_kernels`` mints the fused tree walk of Babbush et al.
+(`arXiv:1805.03662`, Fig. 7) — apply a per-address body exactly when the
+address register equals ``k`` — and ``QROM`` builds coherent
+classical-table lookups on top of it, either as the plain
+unary-iteration SELECT or as the SELECT-SWAP / QROAM construction of
+Low, Kliuchnikov and Schaeffer (`arXiv:1812.00954`).
+
+Everything here is strictly unitary: the papers' headline Toffoli counts
+assume measurement-based ancilla uncomputation, which this library does
+not use (primitives must stay statevector-testable and
+inverse-composable); the module docstrings state the coherent costs the
+minted kernels actually have, and the resource tests pin them against
+the compiler.
+
+Import the subpackage directly (``from cudaq_algorithms.primitives
+import QROM``); nothing here is re-exported from the package root.
+"""
+
+from ._qrom import QROM
+from ._unary_iteration import UnaryIterationKernels, unary_iteration_kernels
+
+__all__ = [
+    "QROM",
+    "UnaryIterationKernels",
+    "unary_iteration_kernels",
+]
--- a/python/cudaq_algorithms/primitives/_qrom.py
+++ b/python/cudaq_algorithms/primitives/_qrom.py
@@ -1,0 +1,326 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""QROM: coherent classical-table lookup, two constructions in one surface.
+
+``QROM(data, address_bits, output_bits)`` mints the lookup
+``|k>|y> -> |k>|y XOR data[k]>`` behind one kernel signature
+``(address: qview, ladder: qview, output: qview)`` — all little-endian
+(qubit 0 = LSB, ``docs/conventions.md``), ``ladder`` being ``num_ladder``
+clean ancillas (|0> in, |0> out). Addresses ``k >= len(data)`` read as
+zero. Two constructions sit behind ``variant=``:
+
+- ``"select"`` — the plain unary-iteration QROM (Babbush et al.,
+  `arXiv:1805.03662`): one fused tree walk over the address register
+  whose per-address body CNOTs the set bits of ``data[k]`` into the
+  output. ``num_ladder = address_bits``.
+- ``"select_swap"`` — SELECT-SWAP / QROAM (Low, Kliuchnikov, Schaeffer,
+  `arXiv:1812.00954`): the table is split into ``ceil(N / B)`` blocks of
+  ``B = block_size`` entries; one unary-iteration walk over the *block
+  index* (the high ``address_bits - log2(B)`` address bits) writes the
+  visited block's ``B`` entries into ``B`` clean ``output_bits``-wide
+  block registers, a swap network controlled on the low ``log2(B)``
+  address bits routes the selected entry's register into block slot 0,
+  and CNOTs copy slot 0 into the output. The routing and the writes are
+  then undone unitarily. ``num_ladder = (address_bits - log2(B)) +
+  B * output_bits`` (walk lines first, then the block registers).
+- ``"auto"`` (default) — price both variants (and every power-of-two
+  ``block_size``) with the exact Toffoli model below and take the
+  cheapest, preferring ``"select"`` on ties (fewer ancillas).
+
+Toffoli accounting (documented contract, pinned by the resource tests;
+``W(n, m)`` is the fused walk count from
+:mod:`cudaq_algorithms.primitives._unary_iteration`, ``3 * 2^n / 2 - 5``
+on full uncontrolled trees):
+
+- ``"select"``: ``W(address_bits, len(data))``.
+- ``"select_swap"``: ``2 * W(address_bits - log2(B), ceil(len(data)/B))
+  + 2 * output_bits * (B - 1)``. Each controlled register swap is
+  ``output_bits`` Fredkins, each Fredkin one Toffoli plus two CNOTs; the
+  binary routing network is ``B - 1`` register swaps per direction, and
+  both the routing and the block writes run twice (compute + unitary
+  uncompute).
+
+The papers' headline counts (``N - 1`` lookup Toffolis for SELECT,
+``ceil(N/B) + output_bits * (B - 1)`` for QROAM) price ancilla
+uncomputation at zero via measurement-and-fixup. These primitives stay
+strictly unitary (house rules: statevector-testable, inverse-composable,
+no mid-circuit measurement), so the coherent counts above are what the
+minted kernels actually cost; the default ``block_size`` optimizes the
+coherent model (``~ sqrt(3 N / (2 output_bits))`` rather than the
+papers' ``~ sqrt(N / output_bits)``).
+
+Both variants are exactly their own inverse: the write phase is X-only,
+so the ``"select"`` walk XORs the same table twice, and the
+``"select_swap"`` sandwich ``U = W S C S^-1 W`` squares to identity
+(``W`` and ``C`` are involutions and the copy commutes with itself
+through the routing). Apply ``kernel()`` again to uncompute; the
+property is pinned by ``tests/python/test_primitives_qrom.py``.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+
+from ._unary_iteration import (_OP_BODY_X, _OP_CCX, _OP_CCX_ADDR_ADDR,
+                               _OP_CCX_CTRL, _OP_CX_ADDR_ADDR,
+                               _OP_CX_ADDR_LADDER, _OP_CX_LADDER_LADDER,
+                               _OP_CX_LADDER_TARGET, _OP_X_ADDR,
+                               _TOFFOLI_OPCODES, _emit_walk, _mint_interpreter,
+                               _walk_toffoli_count, unary_iteration_kernels)
+
+__all__ = ["QROM"]
+
+_VARIANTS = ("auto", "select", "select_swap")
+
+# Walk opcodes whose (a, b) operands index the address register, used to
+# shift the block-index walk onto the high address bits.
+_ADDRESS_OPERANDS = {
+    _OP_X_ADDR: (0, ),
+    _OP_CX_ADDR_LADDER: (0, ),
+    _OP_CCX: (1, ),
+    _OP_CCX_CTRL: (1, ),
+    _OP_CX_ADDR_ADDR: (0, 1),
+    _OP_CCX_ADDR_ADDR: (0, 1),
+}
+
+
+def _select_swap_cost(address_bits: int, num_entries: int, output_bits: int,
+                      block_size: int) -> int:
+    """Documented Toffoli price of the select_swap variant."""
+    high_bits = address_bits - (block_size.bit_length() - 1)
+    num_blocks = -(-num_entries // block_size)
+    return (2 * _walk_toffoli_count(high_bits, num_blocks) + 2 * output_bits *
+            (block_size - 1))
+
+
+class QROM:
+    """Coherent lookup of a classical integer table (see module docstring).
+
+    Parameters
+    ----------
+    data
+        The table: non-negative integers, one per address, each fitting
+        in ``output_bits`` bits.
+    address_bits
+        Address register width (``len(data) <= 2^address_bits``).
+    output_bits
+        Output register width.
+    variant
+        ``"auto"`` (default), ``"select"`` or ``"select_swap"`` — see the
+        module docstring for the constructions and their exact prices.
+    block_size
+        SELECT-SWAP block size: a power of two in
+        ``[2, 2^(address_bits - 1)]``. Only valid with
+        ``variant="select_swap"``; ``None`` there picks the power of two
+        minimizing the documented Toffoli count.
+
+    The three views of the minted kernel may live anywhere (no contiguity
+    requirement between them), which is what lets callers weave the QROM
+    into their own ancilla layouts. Whatever the variant, the kernel is
+    exactly self-inverse — apply it again to uncompute.
+    """
+
+    def __init__(self,
+                 data: Sequence[int],
+                 address_bits: int,
+                 output_bits: int,
+                 *,
+                 variant: str = "auto",
+                 block_size: int | None = None) -> None:
+        if int(address_bits) != address_bits or address_bits < 1:
+            raise ValueError("address_bits must be a positive integer")
+        if int(output_bits) != output_bits or output_bits < 1:
+            raise ValueError("output_bits must be a positive integer")
+        address_bits = int(address_bits)
+        output_bits = int(output_bits)
+        entries = [int(v) for v in data]
+        if len(entries) == 0:
+            raise ValueError("data must be non-empty")
+        if any(int(v) != v for v in data):
+            raise ValueError("data entries must be integers")
+        if len(entries) > (1 << address_bits):
+            raise ValueError(
+                f"data has {len(entries)} entries but address_bits="
+                f"{address_bits} addresses only {1 << address_bits}")
+        if any(v < 0 for v in entries):
+            raise ValueError("data entries must be non-negative")
+        if max(entries) >= (1 << output_bits):
+            raise ValueError(
+                f"data entry {max(entries)} does not fit in output_bits="
+                f"{output_bits} bits (max representable "
+                f"{(1 << output_bits) - 1})")
+        if variant not in _VARIANTS:
+            raise ValueError(
+                f"variant must be one of {_VARIANTS}, got {variant!r}")
+        if block_size is not None:
+            if variant != "select_swap":
+                raise ValueError(
+                    "block_size is only valid with variant='select_swap', "
+                    f"got variant={variant!r}")
+            if (int(block_size) != block_size or block_size < 2
+                    or block_size & (block_size - 1) != 0):
+                raise ValueError("block_size must be a power of two >= 2, got "
+                                 f"{block_size}")
+            block_size = int(block_size)
+            if block_size > (1 << (address_bits - 1)):
+                raise ValueError(
+                    f"block_size {block_size} needs at least one block-"
+                    f"index address bit: max is 2^(address_bits - 1) = "
+                    f"{1 << (address_bits - 1)}")
+
+        select_cost = _walk_toffoli_count(address_bits, len(entries))
+        swap_candidates = [1 << s for s in range(1, address_bits)
+                           ] if block_size is None else [block_size]
+        swap_costs = {
+            size: _select_swap_cost(address_bits, len(entries), output_bits,
+                                    size)
+            for size in swap_candidates
+        }
+        if variant == "auto":
+            best_size = min(swap_costs,
+                            key=lambda size: (swap_costs[size], size),
+                            default=None)
+            if best_size is None or select_cost <= swap_costs[best_size]:
+                variant = "select"
+                block_size = None
+            else:
+                variant = "select_swap"
+                block_size = best_size
+        elif variant == "select_swap":
+            block_size = min(swap_costs,
+                             key=lambda size: (swap_costs[size], size))
+
+        self._data = tuple(entries)
+        self._num_address = address_bits
+        self._output_bits = output_bits
+        self._variant = variant
+        self._block_size = block_size
+        if variant == "select":
+            self._build_select()
+        else:
+            self._build_select_swap()
+
+    def _build_select(self) -> None:
+        entries = self._data
+        output_bits = self._output_bits
+
+        def body(k: int) -> list[tuple[str, int]]:
+            return [("x", t) for t in range(output_bits)
+                    if (entries[k] >> t) & 1]
+
+        # X-only body: the walk is an involution, skip the adjoint mint.
+        walk = unary_iteration_kernels(self._num_address,
+                                       len(entries),
+                                       body,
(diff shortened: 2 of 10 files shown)


python/cudaq_algorithms/primitives/_unary_iteration.py
```diff
@@ -0,0 +1,697 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""Unary iteration: apply a per-address body once for each address value.
```
smcardleQ, line 3: The current explanation is quite wordy, and does not explain functionality in standard quantum computing terms. I would say something along the lines of "Implements unary iteration from Babbush et al (Ref), without using measurement-based uncomputation of temporary logical ANDs. This corresponds to the implementation in Childs et al (Ref). "

In general, I think we should try to write docstrings to be as agent-friendly as possible, assuming that the future library will be consumed primarily by agents. To avoid filling up the context windows, we should have much shorter docstrings, which point to external references that explain the code.

python/cudaq_algorithms/primitives/_unary_iteration.py
```diff
@@ -0,0 +1,697 @@
...
+CNOT-conjugated address wires.
+
+Toffoli cost (unitary-uncompute accounting — see below): a full tree of
+``N = 2^num_address_bits`` addresses costs exactly ``3 N / 2 - 5``
+Toffolis uncontrolled (``N >= 8``; 2 at ``N = 4``, 0 at ``N = 2``) and
+``3 N / 2 - 1`` controlled. Partial trees (``num_items < N``) cost less;
+the emitter reports the exact number as ``toffoli_count``. The paper's
+headline count for unary iteration is ``N - 1`` Toffolis, but that
+figure prices the AND-ancilla *uncomputation* at zero via
+measurement-and-fixup. This library keeps every primitive strictly
```
smcardleQ: I would not say "The paper's headline count is N-1 Toffs" - it is ambiguous what paper this refers to. It suffices to state the current costs, without MBU. Likewise, I don't think it is necessary to state this circuit is from exhaustive search - it is just the Childs et al circuit.

python/cudaq_algorithms/primitives/_unary_iteration.py
```diff
@@ -0,0 +1,697 @@
...
+  — walk with the body applied at each active address (the ``work`` view
+  appears only when ``num_work > 0``).
+- with ``controlled=True``: the same signatures with a leading
+  ``control: qview``, a one-qubit view rooting the tree — the whole walk
+  acts as the identity when the control is |0>. (A separate one-qubit
+  view, not a leading qubit of a combined register, so callers never mix
+  a qview with a bare qubit in one control set. Free body gates still
+  execute at control |0>, but their conjugation pairing cancels them.)
+
+``kernel_adj`` is the hand-written inverse (``cudaq.adjoint`` is
```
smcardleQ, line 95: If we restrict to Pauli (or controlled Pauli) gates, such that we implement \sum_x |x><x| \otimes P_x , then we do not need to reverse the gate ordering. Each Pauli only acts in the branch where the control qubits are in |x>. Hence, for these gates, the unary iteration inverse is equal to its forward.
The exception is if P_x is replaced by general U_x, such that U_x^\dag =/= U_x.

In fact, there are some circuits where we require SELECT to be self-inverse. So it might be helpful to have the option to choose between inverse_SELECT = Reverse/not reversed ordering, in cases where the not reversed ordering is still a valid inverse.

python/cudaq_algorithms/primitives/_unary_iteration.py
```diff
@@ -0,0 +1,697 @@
...
+(X/CX/CCX and controlled Paulis), so the inverse is the same interpreter
+over the reversed instruction list. When the body gates at every address
+mutually commute and square to identity (e.g. the X-only QROM write), the
+walk itself is an involution and ``include_adjoint=False`` skips minting
+the redundant inverse.
+
+The walk circuit
+----------------
+
+Level ``r`` (1-based, from the root) examines address bit
```
smcardleQ: The following description was hard to parse. I think it suffices to point readers to the appendix of Childs et al - the diagram is more clear than the description here, and phrases like "guard", "reclamp" are not standard.

python/cudaq_algorithms/primitives/_unary_iteration.py
```diff
@@ -0,0 +1,697 @@
...
+        if n - t == 1 and not controlled:  # unguarded root crossing
+            total += max(2 * t - 3, 0)
+        else:
+            total += 2 * t - 1
+    return total
+
+
+def _emit_walk(num_address_bits: int, num_items: int, controlled: bool,
+               body: Callable[[int], Sequence[tuple[str, int]]]) -> list:
+    """Flatten the fused tree walk into (opcode, a, b, c) instructions."""
```
smcardleQ, line 333: Explain what body stores in docstring? It seems that body contains the payload list of operations of SELECT. This wasn't clear to me from first reading.

python/cudaq_algorithms/primitives/_unary_iteration.py
```diff
@@ -0,0 +1,697 @@
...
+        emit_body(n - 1, k + 1)
+
+    # Unwind from the last visited leaf.
+    last = num_items - 1
+    for level in range(n, 0, -1):
+        set_line(level, ((last >> wire(level)) & 1) == 1)
+    return ops
+
+
+def unary_iteration_kernels(
```
smcardleQ, line 488: This function looks good to me. However, the core QROM walk is hidden in _emit_walk(), which I am finding hard to parse.

python/cudaq_algorithms/primitives/_unary_iteration.py
```diff
@@ -0,0 +1,697 @@
...
+        if t == 0:
+            continue
+        if n - t == 1 and not controlled:  # unguarded root crossing
+            total += max(2 * t - 3, 0)
+        else:
+            total += 2 * t - 1
+    return total
+
+
+def _emit_walk(num_address_bits: int, num_items: int, controlled: bool,
```
smcardleQ: Are lines 268-408 all one function, _emit_walk() ?
There are a number of functions defined inside, which make it hard to parse and see what is going on. My understanding is that this function is responsible for generating the QROM walk as a list of instructions? In particular, lines 350 - 408 generate the QROM instructions?

kvmto:
Scott, looks good!
The core unary-iteration and QROM coverage looks cool! 
Before merging, could we add:

1. A test that applies `cudaq.control` directly to a generated kernel,`controlled=True` exercises a different path.
2. Behavioral tests for the documented extended body operations (`and_tt`, `and_wt`, `copy_tw`, `x_w`, `z_w`, `sign`, and free gates).

These are advertised capabilities but currently appear untested.
Maybe codex can do it :D

python/cudaq_algorithms/primitives/_unary_iteration.py
```diff
@@ -0,0 +1,697 @@
...
+    emit_body(n - 1, 0)
+
+    for k in range(num_items - 1):
+        t = _trailing_ones(k)
+        if t == 0:
+            # Sibling flip at the deepest level: left XOR right = parent.
+            if n == 1:
+                root_flip()
+            else:
+                emit(_OP_CX_LADDER_LADDER, n - 2, n - 1)
```
smcardleQ: I think it is a bit confusing that we deal with both the root_flip() [top level of tree] and the sibling leaf transition [bottom level of the tree, done here with CNOT(L_{n-2}, L{n-1}))] in the same block of code.

We could consider stripping out n==1 case into its own separate QROM block, because it is so simple (literally two c-payload gates, in the uncontrolled case).

That said, it is nice that the current code template fits all n values, so I am ok to leave it be. 
Perhaps rename "emit(_OP_CX_LADDER_LADDER, n - 2, n - 1)" here as leaf_transition(), defined separately above?

python/cudaq_algorithms/primitives/_unary_iteration.py
```diff
@@ -0,0 +1,697 @@
...
+        elif n - t == 1 and not controlled and t >= 2:
```
smcardleQ: This feels like a very particular case, and the code in lines 369-376 is almost hardcoded. Suggest we factor this into a separate short function that explains what it does in its name, and call it here?

python/cudaq_algorithms/primitives/_unary_iteration.py
```diff
@@ -0,0 +1,697 @@
...
+def _emit_walk(num_address_bits: int, num_items: int, controlled: bool,
+               body: Callable[[int], Sequence[tuple[str, int]]]) -> list:
+    """Flatten the fused tree walk into (opcode, a, b, c) instructions."""
+    n = num_address_bits
+    ops: list[tuple[int, int, int, int]] = []
+
+    def emit(opcode: int, a: int = 0, b: int = 0, c: int = 0) -> None:
+        ops.append((opcode, a, b, c))
+
+    def emit_body(line: int, k: int) -> None:
```
smcardleQ, line 340: My understanding is that emit body will take the kth operator stored inside the 'body' list, which is a list of gates. It will then apply the gates onto their specified qubits. E.g. X0 Y2 Z8.

Currently, _BODY_OPCODES is limited to Pauli gates. Do we plan to expand this beyond just Paulis? This is sufficient for QROM, or basic Pauli LCU. But there may be cases where we want to apply operators beyond just Paulis, in SELECT.

python/cudaq_algorithms/primitives/_unary_iteration.py
```diff
@@ -0,0 +1,697 @@
...
+            emit(_OP_CX_ADDR_ADDR, w1, w2)
+            emit(_OP_CX_ADDR_ADDR, w1, w3)
+            emit(_OP_CCX_ADDR_ADDR, w2, w3, 2)
+            emit(_OP_CX_ADDR_ADDR, w1, w3)
+            emit(_OP_CX_ADDR_ADDR, w1, w2)
+            emit(_OP_CX_ADDR_LADDER, w1, 1)
+            emit(_OP_CX_ADDR_LADDER, w2, 1)
+            root_flip()
+            for level in range(4, n + 1):
+                set_line(level, False)  # reclamp against new parents
```
smcardleQ, line 459: This feels like a very particular case, and the code in lines 369-376 is almost hardcoded. Suggest we factor this into a separate short function that explains what it does in its name, and call it here?

python/cudaq_algorithms/primitives/_unary_iteration.py
smcardleQ: I think it is a bit confusing that we deal with both the root_flip() [top level of tree] and the sibling leaf transition [bottom level of the tree, done here with CNOT(L_{n-2}, L{n-1}))] in the same block of code.

We could consider stripping out n==1 case into its own separate QROM block, because it is so simple (literally two c-payload gates, in the uncontrolled case).

That said, it is nice that the current code template fits all n values, so I am ok to leave it be.
Perhaps rename "emit(_OP_CX_LADDER_LADDER, n - 2, n - 1)" here as leaf_transition(), defined separately above?

wsttiger:
Both added in ad507a9 (`tests/python/test_primitives_body_gates.py`, 8 tests):

1. `test_external_cudaq_control_matches_builtin_controlled` builds the same walk twice — plain kernel wrapped in `cudaq.control`, and `controlled=True` — and requires both statevectors to equal each other *and* an analytic reference on a superposed control, so the two routes can't agree by being wrong the same way. (They intentionally differ in cost — the built-in route folds the control into the tree at 3N/2−1 Toffolis, the wrap controls every gate — so the resource pins stay on the built-in route only.)
2. One behavioral test per extended gate: each states its expected action as a `bits -> (bits, phase)` function at the active address and checks the full statevector on a phased superposed address, exhaustively over target basis states, with work registers verified back to |0⟩. The free-gate tests also pin the soundness condition (conjugation pairs cancel exactly on inactive addresses), and a resource test holds `toffoli_count` equal to the compiler's count when the body contributes its own Toffolis (`and_tt`).

Full suite: 399 passed / 3 skipped.

tests/python/test_primitives_unary_iteration.py
```diff
@@ -0,0 +1,321 @@
...
+
+
+def test_unary_iteration_adjoint_composes_to_identity():
+    # Non-commuting multi-gate bodies (X, Z, Y mixed) so the walk is NOT
+    # an involution; only the hand-written reversed-instruction inverse
+    # restores the state.
+    def body(k):
+        if k % 2 == 1:
+            return [("x", 0), ("z", 0), ("y", 1)]
+        return [("x", 1)]
```
smcardleQ, line 258: I'm not sure I agree that this test checks what it says it does. As written here, I believe it implements:

SEL = \sum_x |x><x| \otimes P_x

As defined this way, SEL is self-inverse, so reverse-ordering will not matter. I think we need a more rigorous test of explicit inverse, in the case where SEL is not self-inverse. That may require expanding _BODY_OPS to include at least one non-Hermitian operator.

tests/python/test_primitives_unary_iteration.py
```diff
@@ -0,0 +1,321 @@
...
+
+    state = np.array(cudaq.get_state(run))
+    expected = np.zeros(1 << 8, dtype=np.complex128)
+    for address in range(8):
+        expected[address + (1 << 6)] = 1.0 / np.sqrt(8.0)
+    np.testing.assert_allclose(state, expected, atol=1e-12)
+
+
+@pytest.mark.parametrize("control_value", [0, 1])
+def test_unary_iteration_controlled_variant(control_value):
```
smcardleQ: I'd like to see this tested on a superposition of control, not just on c=0, c=1 individually. The former is required to catch phase errors, which could sneak in once we add measurement-based uncomputation.

python/cudaq_algorithms/primitives/__init__.py
```diff
@@ -0,0 +1,30 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""Fault-tolerant circuit primitives: unary iteration and QROM.
+
+``unary_iteration_kernels`` mints the fused tree walk of Babbush et al.
+(`arXiv:1805.03662`, Fig. 7) — apply a per-address body exactly when the
+address register equals ``k`` — and ``QROM`` builds coherent
```
smcardleQ: Should we clarify that it is actually the walk of Childs et al, because we don't use measurement-based uncomputation here?

python/cudaq_algorithms/primitives/_qrom.py
```diff
@@ -0,0 +1,326 @@
...
+        self._num_ladder = walk.num_ladder
+        self._toffoli_count = walk.toffoli_count
+
+    def _build_select_swap(self) -> None:
+        entries = self._data
+        b = self._output_bits
+        size = self._block_size
+        low_bits = size.bit_length() - 1
+        high_bits = self._num_address - low_bits
+        num_blocks = -(-len(entries) // size)
```
smcardleQ: Why is there a double minus sign, inside and outside of parentheses?

python/cudaq_algorithms/primitives/_qrom.py
```diff
@@ -0,0 +1,326 @@
...
+                gates.extend(
+                    ("x", i * b + t) for t in range(b) if (word >> t) & 1)
+            return gates
+
+        walk_ops = _emit_walk(high_bits, num_blocks, False, block_write)
+        # Rebase the walk onto the combined kernel: block-index address
+        # wires sit above the low routing bits, and the walk's "target"
+        # (the block registers) lives in the ladder view after the walk
+        # lines, so the leaf-controlled body X becomes a ladder-to-ladder
+        # CNOT.
```
smcardleQ: I don't understand what this comment is trying to tell me. My understanding of SEL-SWAP is that we should do the following:
1. Run unary iteration on high-bits. The high-bits index blocks 0, 1, ...,  \ceil[N/B]-1, and the unary iteration leaf action writes each block of B entries.
2. Use the SWAP network controlled on the low-bits to coherently route the B entries.

What is meant here by "the walk's target lives in the ladder view, so leaf-controlled body X becomes a ladder-to-ladder CNOT" ? 

I think we should be able to directly reuse the regular unary_iteration primitive.

python/cudaq_algorithms/primitives/_qrom.py
```diff
@@ -0,0 +1,326 @@
...
+        # Fredkins: cswap(c; u, v) = cx(v, u) ccx(c, u, v) cx(v, u).
+        swap_ops = []
+        for s in range(low_bits):
+            for i in range(0, size, 1 << (s + 1)):
+                for t in range(b):
+                    u = high_bits + i * b + t
+                    v = high_bits + (i + (1 << s)) * b + t
+                    swap_ops.append((_OP_CX_LADDER_LADDER, v, u, 0))
+                    swap_ops.append((_OP_CCX, u, s, v))
+                    swap_ops.append((_OP_CX_LADDER_LADDER, v, u, 0))
```
smcardleQ, line 349: As above, I am confused by _OP_CX_LADDER_LADDER. 
I would expect the CNOTs to purely act between the data write blocks, to implement the swapping.

python/cudaq_algorithms/primitives/_unary_iteration.py
wsttiger: You're right on both counts, and on the attribution more than we knew: we checked arXiv:1711.10980 and this walk is the select(V) construction of Appendix G.4, Lemma G.7 — same in-place fused transitions (their Eqs. 179–181 shortcuts), same 3N/2 + O(1) count, no measurement, even the same dirty-target 7-T caveat. The docstring now opens with essentially your sentence (8d5591f) and the exhaustive-search narrative is gone; what remains ours is the presentation, the priced controlled variant (3N/2 − 1), and non-power-of-2 handling. Agreed on short-docstrings-that-cite as house style going forward — the API-contract sections (opcode set, signatures, conventions) stay in-module since agents need those mechanically.

python/cudaq_algorithms/primitives/_unary_iteration.py
wsttiger: Fixed in 8d5591f with the attribution rewrite — costs stated plainly with named citations (Childs et al. for this walk, Babbush et al. for the N − 1 measured figure), exhaustive-search sentence dropped.

python/cudaq_algorithms/primitives/_unary_iteration.py
wsttiger: Agreed on the math: for single-Pauli bodies SELECT is self-inverse and the reversal is unnecessary. That option exists as `include_adjoint=False` — the factory skips minting the redundant inverse and the caller reuses `kernel` as its own adjoint (documented at the factory docstring: 'pass False when the walk is known to be an involution, e.g. the X-only QROM write'). QROM uses exactly this. For general non-Hermitian U_x, the reversed-tape `kernel_adj` stays the safe default.

python/cudaq_algorithms/primitives/_unary_iteration.py
wsttiger: Done in 8d5591f — the `body` docstring now leads with your framing: it is the payload table of the SELECT, mapping each address k to the gate list U_k, so the minted kernel implements sum_k |k><k| ⊗ U_k; plus a note that body runs at factory time only (the gates are baked in, body never executes at circuit time).

python/cudaq_algorithms/primitives/_unary_iteration.py
wsttiger: Partially addressed in 8d5591f: the transition loop now reads as three named cases — `sibling_flip`, `unguarded_root_crossing`, and the guarded fused transition — each with a docstring stating its algebraic identity. And with the Childs et al. attribution in place, the walk now has a citable external explanation (their Appendix G.4 figures) rather than only in-code prose.

python/cudaq_algorithms/primitives/_unary_iteration.py
wsttiger: Yes — that expansion is already in this PR: the extended body vocabulary (`and_tt`, `and_wt`, `copy_tw`, `x_w`, `z_w`, `sign`, plus the free conjugation gates) supports multi-controlled bit flips and phases via work qubits, with behavioral tests in tests/python/test_primitives_body_gates.py (ad507a9). What it deliberately does not yet include is continuous gates — leaf-controlled rotations, the data-looked-up-angles SELECT that THC needs. That's planned as a follow-up when THC pays for it, since it changes the resource story (rotations don't pin to a Toffoli count).

python/cudaq_algorithms/primitives/_unary_iteration.py
wsttiger: Done in 8d5591f — it is now `unguarded_root_crossing()`, with a docstring stating what makes the case special (no guard line above the root, so the level-2 difference is fully linear and level 3 costs one Toffoli on CNOT-conjugated wires).

python/cudaq_algorithms/primitives/_unary_iteration.py
wsttiger: Done in 8d5591f via a small generalization of your suggestion: `sibling_flip(level)` covers every level (left XOR right = parent, one CNOT), with level == n being your leaf_transition and level == 1 degenerating to root_flip since there is no parent line. The n == 1 case then falls out of the same helper rather than needing its own block.

tests/python/test_primitives_unary_iteration.py
wsttiger: The bodies here are actually multi-gate and non-commuting — U_odd = Y1·Z0·X0, which is non-Hermitian, so this SELECT is not self-inverse: (Z0 X0)^2 = −I, and a kernel_adj that secretly replayed the forward tape would leave a −1 phase on the odd addresses, visible against the superposed address register. So the reversal is genuinely exercised. Your underlying concern is fair though: nothing *pinned* the non-involution premise, so the test could have silently degraded if someone later simplified the body. 8d5591f adds the explicit premise assertion — forward∘forward is checked to be far from identity before forward∘adjoint is required to equal it.

tests/python/test_primitives_unary_iteration.py
wsttiger: Done twice over: this test now runs ry(θ) with θ ∈ {0, π, 0.7} — the classical branches plus genuine superposition, against the amplitude-weighted expected state (8d5591f) — and test_external_cudaq_control_matches_builtin_controlled (ad507a9) already runs both control routes on ry(0.7) against an analytic reference. Fully agreed on the motivation: the branch-relative phase error is exactly the failure mode measurement-based uncomputation would introduce, and these are the tests that would catch it when that track is promoted.

python/cudaq_algorithms/primitives/__init__.py
wsttiger: Yes — done in 58dae3a (package docstring) and 8d5591f (module docstring), and independently confirmed against the paper: Appendix G.4, Lemma G.7 is this walk — 1.5·2^w − 4 Toffolis, fused in-place transitions, measurement-free.

python/cudaq_algorithms/primitives/_qrom.py
wsttiger: It was ceiling division by double negation (-(-a // b) = ceil(a/b) for positive ints). Replaced with the additive idiom (len + size − 1) // size plus a comment in 58dae3a — same result, no double take.

python/cudaq_algorithms/primitives/_qrom.py
wsttiger: Your two-step description is exactly what this implements — walk over the high (block-index) bits whose leaf action writes each block, then the low-bit-controlled swap network. The confusing part was the comment, not the construction, and 58dae3a rewrites it. On direct reuse: we reuse the walk *emitter* (_emit_walk) rather than a minted walk kernel because the whole lookup must be one flat instruction tape — CUDA-Q kernels cannot call kernels — so the walk's instructions get rebased into the combined kernel's wire layout. In that layout the block registers ride in the 'ladder' *view* after the walk lines ('ladder' names the wire bundle, not a role), which is why the leaf-controlled block-write X lands as a CNOT between two ladder-view wires. The register layout is documented in _unary_iteration's docstring ('select_swap register layout', aac87c8).

python/cudaq_algorithms/primitives/_qrom.py
wsttiger: They are purely between the data write blocks — u and v in the Fredkin decomposition are block-register wires, which live in the ladder *view* (see the rebase note above this block), so the routing network's own CNOTs carry the _OP_CX_LADDER_LADDER opcode. The opcode name refers to the wire bundle, not the gate's role; 58dae3a adds a note at the routing network saying exactly this.

wsttiger:
Added one commit (268462c): CUDA-Q 0.16.0 released without the fix for cuda-quantum#5280, so that compiler hangs on this feature's 1-bit-address lookups (a classically-dead dispatch arm in the minted interpreters gets CSE'd into ill-formed IR). This commit prunes unused dispatch arms at mint time, removing the trigger entirely — no public API change, and the full suite is verified green against released cudaq 0.15.1 and against the 0.16.0 wheel. Details in the commit message.

kvmto (approved):
LGTM!

PR #42, opened 2026-08-26
Test PRs against released cudaq; from-source becomes a manual probe

wsttiger:
PR CI has been building cuda-quantum from source at the pinned `.cudaq_version` — a moving development snapshot our wheel explicitly does not support (`cudaq >= 0.15.0, < 0.16`). Two failure modes in one week came from that mismatch: a pin behind upstream's nanobind fix could not build (#41), and the current pin ahead of it hangs the 0.16-dev compiler on legal 0.15 kernels (NVIDIA/cuda-quantum#5280, reproducer included there).

This PR makes CI test what we ship for:

- **Algorithms lanes**: new `cudaq_source` input (default `pip`) — PRs `pip install "cudaq>=0.15.0,<0.16"` and test against it; patch releases inside the range are picked up automatically. The test script gains `--pip-cudaq` so a container-baked `/usr/local/cudaq` cannot shadow the pip install on `PYTHONPATH`.
- **Wheel lanes**: PRs run `cudaq_wheels: PyPI` — the exact mode the v0.1.0 release validated end to end.
- **The from-source path is not removed**: `workflow_dispatch` with `cudaq_source: source` / `cudaq_wheels: Custom` runs the old pipeline deliberately, as a forward-compatibility probe where a dev-cudaq failure is a finding (like #5280 was) rather than a PR blocker.

Note this PR's own CI is the validation: its lanes run the new pip path. The bench targets used for resource counting (`compiler-bench-ftqc-logical`) are present in released 0.15.1, so nothing test-side loses capability.

### DIFF
--- a/.github/workflows/lib_algorithms.yaml
+++ b/.github/workflows/lib_algorithms.yaml
@@ -119,7 +119,7 @@ jobs:
           python3 -m pip install numpy scipy pytest pyscf qsppack
 
       - name: Install released cudaq
-        if: inputs.cudaq_source != 'source'
+        if: inputs.cudaq_source == 'pip'
         # The range mirrors pyproject's declared dependency; patch
         # releases inside it are tested automatically. The cudaq
         # resolver picks its own cuda-quantum-cuNN on these CPU-only
--- a/.github/workflows/lib_algorithms.yaml
+++ b/.github/workflows/lib_algorithms.yaml
@@ -3,6 +3,17 @@ name: Algorithms lib
 on:
   workflow_dispatch:
     inputs:
+      cudaq_source:
+        type: choice
+        description: >
+          cudaq to test against: pip = the released cudaq the wheel
+          declares support for; source = build the pinned .cudaq_version
+          from source (forward-compatibility probing).
+        options:
+          - pip
+          - source
+        default: pip
+        required: false
       cache_key_suffix:
         type: string
         description: Optional suffix to append to CUDA-Q cache keys.
@@ -20,6 +31,10 @@ on:
         required: false
   workflow_call:
     inputs:
+      cudaq_source:
+        type: string
+        default: pip
+        required: false
       cache_key_suffix:
         type: string
         default: ''
@@ -76,6 +91,7 @@ jobs:
       # ========================================================================
 
       - name: Get required CUDAQ version
+        if: inputs.cudaq_source == 'source'
         id: get-cudaq-version
         uses: ./.github/actions/get-cudaq-version
         with:
@@ -83,6 +99,7 @@ jobs:
           ref: ${{ inputs.cudaq_ref }}
 
       - name: Get CUDAQ build
+        if: inputs.cudaq_source == 'source'
         uses: ./.github/actions/get-cudaq-build
         with:
           repo: ${{ steps.get-cudaq-version.outputs.repo }}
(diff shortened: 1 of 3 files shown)


.github/workflows/pr_workflow.yaml
anjbur: This should be removed from the filter list, since the `build-wheels` is hardcoded to use PyPI rather than build from source

.github/workflows/lib_algorithms.yaml
```diff
@@ -101,9 +118,23 @@ jobs:
         run: |
           python3 -m pip install numpy scipy pytest pyscf qsppack
 
+      - name: Install released cudaq
+        if: inputs.cudaq_source != 'source'
```
anjbur:
```suggestion
        if: inputs.cudaq_source == 'pip'
```
nit: checking for pip is a bit clearer to read and future-proofs against any potential changes to the `cudaq_source` options

.github/workflows/pr_workflow.yaml
anjbur, line 121: The gate for `build-cudaq` can be removed, since this job always uses pip as the `cudaq_source`. It'll also need to be removed on lines 104 and 107

anjbur (commented):
`.cudaq_version` is still wired into the testing structure of `pr_workflow.yaml`. The job `build-wheels` has a trigger for changes to it (line 79), despite being hard-coded to use PyPI instead of the file-pinned version. The job `build-cudaq` has a similar trigger and does use the pinned commit, but `build-algorithms` still tests against the `pip`-installed released cudaq, despite being gated on the `build-cudaq` job. All that to say, a pin bump will trigger builds, but it won't run any tests against the changed pin.

I'd suggest creating two new jobs that build the wheels and algorithms library from the cudaq source in `.cudaq_version` (rather than from pip/PyPI) and have those jobs only be triggered on changes to `.cudaq_version`. That way we can still have effective PR testing when the pinned version is bumped, but the majority of PRs will run against the previously released version. If the different versions of wheel and algorithm build have differentiating names, then it'll also make it more clear to the PR author what path is being tested.

.github/workflows/pr_workflow.yaml
wsttiger: Removed — the PyPI wheel lane can't see the pin, so the trigger was stale. Pin changes now route to the new `build-wheels-source` job instead (fb90f3c).

.github/workflows/lib_algorithms.yaml
wsttiger: Done — the positive condition reads better and won't silently match a future third option.

.github/workflows/pr_workflow.yaml
wsttiger: Removed the `needs` edge and both result/output conditions; the pip lane now depends only on `check-changes`. The pin-bump trigger that used to piggyback on this job moved to `build-algorithms-source`, which is the job that actually consumes the build.

wsttiger:
Good catch on the pin-bump gap — you're right that a `.cudaq_version` change triggered builds but tested nothing against the changed pin. Implemented your suggestion in fb90f3c:

- New `build-algorithms-source` (Algorithms, `cudaq_source: source`) and `build-wheels-source` (Wheels, `cudaq_wheels: Custom` — that mode already builds the CUDA-Q wheels from the pinned commit), both gated on the `build-cudaq` path filter, so they run exactly when `.cudaq_version` or the cudaq build plumbing changes.
- Ordinary PRs keep running only the released-cudaq lanes.
- The job names show up in the checks list as "Algorithms (pinned cudaq, from source)" / "Wheels (pinned cudaq, from source)", so it's visible which path a PR exercised.

One honest caveat: this PR doesn't touch `.cudaq_version`, so its own CI exercises the pip lanes but not the two new source jobs — they get their first real run on the next pin bump. If you'd rather see them proven before merge, I can open a scratch PR with a trivial `.cudaq_version` touch to light them up.

anjbur (approved):
LGTM!

PR #43, opened 2026-08-28
fix: Probe CuPy with a real kernel and collect CuPy DF tests only on nvidia simulator

kaiqiy-nv:
This PR tries to fix [[B] 6682868](https://nvbugspro.nvidia.com/bug/6682868).
It also adds 3 test cases to expose the bug and verify the fix.

In this suggested fix:
We update `cupy_gpu_available()` so a visible device also has to run a tiny compiled kernel (`arange`) before `"auto"` may select CuPy.
In addition, we collect the `[cupy]` double-factorization tests only when `CUDAQ_DEFAULT_SIMULATOR` starts with `nvidia` (with a call-time skip if that probe then fails).
In this case, `backend="auto"` choosing CuPy when the GPU is listed but cannot compute, and CuPy DF tests running (and failing) on the `qpp-cpu` validation leg, can be avoided.

Added test cases:
- `test_cupy_gpu_available_requires_a_working_kernel`:
We install a fake CuPy whose `getDeviceCount()` is 2 but whose `arange` raises, then check that `cupy_gpu_available()` is False so a driver-only / dead-context GPU is not treated as usable.
- `test_auto_uses_numpy_when_cupy_kernel_probe_fails`:
We use the same fake CuPy, drop the size crossovers so `"auto"` would have picked CuPy under the old probe, then run X-DF and C-DF with `backend="auto"` and check that both still return a valid factorization (NumPy path) so the old crash is exposed if the kernel probe is dropped.
- `test_explicit_cupy_backend_errors_when_kernel_probe_fails`:
We use the same fake CuPy and call X-DF with `backend="cupy"`, then check that it raises `RuntimeError` matching `no CuPy/GPU is available` so a hard CuPy request does not silently fall back.

Tests passed on local machine as:
```
============================= test session starts ==============================
platform linux -- Python 3.12.13, pytest-9.1.1, pluggy-1.6.0 -- /root/.miniconda3/envs/algo312/bin/python
rootdir: /CUDA-QX/kaiqiy/cudaq-algorithms
configfile: pyproject.toml
collecting ... 1
collected 4 items

tests/python/test_double_factorization.py::test_cupy_gpu_available_requires_a_working_kernel PASSED [ 25%]
tests/python/test_double_factorization.py::test_auto_uses_numpy_when_cupy_kernel_probe_fails[xdf] PASSED [ 50%]
tests/python/test_double_factorization.py::test_auto_uses_numpy_when_cupy_kernel_probe_fails[cdf] PASSED [ 75%]
tests/python/test_double_factorization.py::test_explicit_cupy_backend_errors_when_kernel_probe_fails PASSED [100%]

============================== 4 passed in 2.04s ===============================
```

### DIFF
--- a/python/cudaq_algorithms/double_factorization/_backend.py
+++ b/python/cudaq_algorithms/double_factorization/_backend.py
@@ -61,11 +61,13 @@ def resolve_backend(backend: str = "auto",
                     gpu_min_size: int = 0) -> tuple[ArrayModule, str]:
     """Return ``(array_module, name)`` for the requested backend.
 
-    ``"auto"`` selects CuPy when a GPU is available *and* the problem is large
-    enough to amortize GPU launch/sync overhead -- i.e. ``problem_size`` (the
-    orbital count ``n``) is at least ``gpu_min_size`` -- otherwise NumPy. With no
-    ``problem_size`` hint it keeps the legacy behavior: CuPy whenever a GPU is
-    present. ``"cupy"`` / ``"numpy"`` force the backend regardless of size.
+    ``"auto"`` selects CuPy when the problem is large enough to amortize GPU
+    launch/sync overhead -- i.e. ``problem_size`` (the orbital count ``n``) is
+    at least ``gpu_min_size`` -- *and* a GPU is available; otherwise NumPy. The
+    size check is evaluated first so a below-threshold problem does not create
+    a CUDA context. With no ``problem_size`` hint it keeps the legacy behavior:
+    CuPy whenever a GPU is present. ``"cupy"`` / ``"numpy"`` force the backend
+    regardless of size.
     """
     if backend == "numpy":
         return np, "numpy"
@@ -75,8 +77,8 @@ def resolve_backend(backend: str = "auto",
                                "CuPy/GPU is available.")
         return _cupy, "cupy"
     if backend == "auto":
-        if cupy_gpu_available() and (problem_size is None
-                                     or problem_size >= gpu_min_size):
+        if ((problem_size is None or problem_size >= gpu_min_size)
+                and cupy_gpu_available()):
             return _cupy, "cupy"
         return np, "numpy"
     raise ValueError(f"unknown backend '{backend}'; expected 'auto', 'cupy', "
--- a/python/cudaq_algorithms/double_factorization/_backend.py
+++ b/python/cudaq_algorithms/double_factorization/_backend.py
@@ -75,8 +75,10 @@ def resolve_backend(backend: str = "auto",
         return np, "numpy"
     if backend == "cupy":
         if not cupy_gpu_available():
-            raise RuntimeError("the 'cupy' backend was requested but no "
-                               "CuPy/GPU is available.")
+            raise RuntimeError(
+                "the 'cupy' backend was requested but CuPy/GPU is not "
+                "usable (not installed, no device, or the kernel probe "
+                "failed).")
         return _cupy, "cupy"
     if backend == "auto":
         if ((problem_size is None or problem_size >= gpu_min_size)
--- a/python/cudaq_algorithms/double_factorization/_backend.py
+++ b/python/cudaq_algorithms/double_factorization/_backend.py
@@ -43,14 +43,16 @@ def cupy_gpu_available() -> bool:
 
     ``getDeviceCount`` only talks to the driver, so a driver-only install
     (no NVRTC / CUDA toolkit) would otherwise look usable and then fail on
-    the first compiled kernel. The tiny ``arange`` forces that compile.
+    the first compiled kernel. The tiny ``arange`` forces that compile, and
+    ``.sum()`` into a host ``float`` waits for the device so a fault that
+    only appears at synchronization is not missed.
     """
     if _cupy is None:
         return False
     try:
         if _cupy.cuda.runtime.getDeviceCount() <= 0:
             return False
-        _cupy.arange(1) + 1
+        float((_cupy.arange(1) + 1).sum())
         return True
     except Exception:  # pragma: no cover - driver/runtime/NVRTC issues
         return False
--- a/python/cudaq_algorithms/double_factorization/_backend.py
+++ b/python/cudaq_algorithms/double_factorization/_backend.py
@@ -39,12 +39,20 @@ AUTO_GPU_MIN_ORBITALS_EXPLICIT = 56
 
 
 def cupy_gpu_available() -> bool:
-    """Return True when CuPy is importable and at least one GPU is visible."""
+    """Return True when CuPy can run a kernel on a visible GPU.
+
+    ``getDeviceCount`` only talks to the driver, so a driver-only install
+    (no NVRTC / CUDA toolkit) would otherwise look usable and then fail on
+    the first compiled kernel. The tiny ``arange`` forces that compile.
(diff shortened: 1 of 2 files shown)


tests/python/test_double_factorization.py
```diff
@@ -13,12 +13,25 @@
...
 df = algorithms.double_factorization
 
-# Run every test on the NumPy backend, plus the CuPy (GPU) backend when present.
+# NumPy always. CuPy only on the GPU simulator leg (nvidia*), never under
+# qpp-cpu: these tests are classical linear algebra, and a dead GPU must not
+# be reported as a CPU-simulator failure.
 _BACKENDS = ["numpy"]
-if df.cupy_gpu_available():
+if os.environ.get("CUDAQ_DEFAULT_SIMULATOR", "qpp-cpu").startswith("nvidia"):
     _BACKENDS.append("cupy")
```
wsttiger, line 24: This gate silently drops the 14 `[cupy]` parametrizations on any GPU dev box where `CUDAQ_DEFAULT_SIMULATOR` is unset — no skip, no reason, they just aren't collected (verified: 44 tests on main vs 34 here on a working-GPU box). CI is fine (the GPU wheel legs set `nvidia-fp64`, which `startswith("nvidia")` matches, and the CPU legs never had cupy installed), but a developer running `pytest` on a GPU workstation loses coverage without noticing.

There's also a proxy mismatch: these tests are classical linear algebra, so CuPy usability has nothing to do with which CUDA-Q simulator is configured. Since your new autouse fixture already handles the dead-GPU case with a visible call-time skip, could this collect when the probe passes as well?

```python
if (os.environ.get("CUDAQ_DEFAULT_SIMULATOR", "qpp-cpu").startswith("nvidia")
        or df.cupy_gpu_available()):
    _BACKENDS.append("cupy")
```

That keeps the qpp-cpu validation leg clean (probe fails there when no GPU) while preserving dev-box coverage.

python/cudaq_algorithms/double_factorization/_backend.py
```diff
@@ -39,12 +39,20 @@
...
+    the first compiled kernel. The tiny ``arange`` forces that compile.
+    """
     if _cupy is None:
         return False
     try:
-        return _cupy.cuda.runtime.getDeviceCount() > 0
-    except Exception:  # pragma: no cover - driver/runtime issues
+        if _cupy.cuda.runtime.getDeviceCount() <= 0:
+            return False
+        _cupy.arange(1) + 1
```
wsttiger: Nit: the launch is asynchronous and the result is discarded, so a fault that only surfaces at synchronization would slip through the probe. `float((_cupy.arange(1) + 1).sum())` forces the device round-trip at the same cost. (NVRTC/compile failures — the bug class you're fixing — do raise at launch, so this is belt-and-suspenders.)

wsttiger (changes requested):
The core fix is right and the tests are excellent — I reverted `_backend.py` to main while keeping your new tests, and all four fail against the old probe exactly as intended, then pass with the fix restored. The kernel probe is also the correct strength (`arange` exercises the NVRTC compile path; `zeros` would only memset). Two things I'd like addressed before merge, both verified on a live GPU box (2× RTX 6000 Ada, cupy 13.6) and both cheap; details inline.

1. `backend="auto"` now creates a full CUDA context before the size check — measured ~234 ms and ~689 MB of device memory held for the process lifetime, even when auto then picks numpy for a small problem. One-line reorder fixes it.
2. The collection gate silently removes the 14 `[cupy]` parametrizations on any GPU dev box where `CUDAQ_DEFAULT_SIMULATOR` is unset (main runs 44 tests there, this PR runs 34, with no skip marker). CI is unaffected (the GPU wheel legs set `nvidia-fp64`, and cupy is never installed on the CPU legs anyway), but developer-box coverage drops invisibly.

Also verified clean, for the record: no monkeypatch leakage between the fake-CuPy tests and real-GPU tests (48/48 under `CUDAQ_DEFAULT_SIMULATOR=nvidia` with both coexisting); the call-time skip fixture reports a visible skip (34 passed / 14 skipped with a broken-cupy shim under `nvidia-fp64`); non-ImportError failures at cupy import are handled; yapf clean.

---

Two points on lines outside this diff (so no inline anchor):

**`_backend.py:78` (resolve_backend, the `auto` branch)** — Left-to-right evaluation runs the kernel probe before the size check, and unlike the old `getDeviceCount` (a context-free driver call), `arange(1) + 1` initializes a CUDA context: measured 233.8 ms and 689 MB of device memory that stays held for the process lifetime — paid even when the very next check decides the problem is too small and returns numpy. On a shared GPU node, every small-molecule X-DF/C-DF call with `backend="auto"` now pins ~0.7 GB of someone's GPU for nothing.

Suggest swapping the operands so the cheap size check short-circuits:

```python
if (problem_size is None or problem_size >= gpu_min_size) \
        and cupy_gpu_available():
```

**`_backend.py:74-76` (explicit `backend="cupy"` error)** — Nit: with this PR, this error also fires in a new case — CuPy installed, GPU visible, but the kernel probe failed — where "no CuPy/GPU is available" will read as wrong to the user staring at their nvidia-smi output (and the new `test_explicit_cupy_backend_errors_when_kernel_probe_fails` pins the wording). Something like "CuPy/GPU is not usable (not installed, no device, or the kernel probe failed)" would tell them why.

kaiqiy-nv:
Thanks @wsttiger .
I have addressed all the comments in the 4 commits above.
Tests passed on local machine as:
```
============================= test session starts ==============================
platform linux -- Python 3.12.13, pytest-9.1.1, pluggy-1.6.0 -- /root/.miniconda3/envs/algo312/bin/python
rootdir: /CUDA-QX/kaiqiy/cudaq-algorithms
configfile: pyproject.toml
collecting ... collected 5 items

tests/python/test_double_factorization.py::test_cupy_gpu_available_requires_a_working_kernel PASSED [ 20%]
tests/python/test_double_factorization.py::test_auto_uses_numpy_when_cupy_kernel_probe_fails[xdf] PASSED [ 40%]
tests/python/test_double_factorization.py::test_auto_uses_numpy_when_cupy_kernel_probe_fails[cdf] PASSED [ 60%]
tests/python/test_double_factorization.py::test_explicit_cupy_backend_errors_when_kernel_probe_fails PASSED [ 80%]
tests/python/test_double_factorization.py::test_auto_backend_is_size_aware PASSED [100%]

============================== 5 passed in 2.20s ===============================
```

wsttiger (approved):
All four points addressed — and I re-verified the two blocking ones by execution on a live GPU box (2x RTX 6000 Ada, cupy 13.6):

1. **No premature CUDA context**: `resolve_backend("auto", problem_size=4, gpu_min_size=56)` now returns numpy in ~0.00 ms with zero GPU memory attached to the process (previously ~234 ms and ~689 MB held for the process lifetime). The `_must_not_probe` monkeypatch test is a stronger guard than I asked for — nice touch; the regression can't come back silently.
2. **Dev-box coverage restored**: with `CUDAQ_DEFAULT_SIMULATOR` unset on a working-GPU machine, the file now collects and passes **48 tests** (the previous revision collected 34; main runs 44) — so this lands with more coverage than main, not less.

The synced probe (`float((arange(1) + 1).sum())`) and the clarified error message both read exactly right. Thanks for the quick turnaround.

PR #45, opened 2026-09-01
fix: Fix formula coefficient in the comment block

kaiqiy-nv:
This PR tries to fix [[B] 6702274](https://nvbugspro.nvidia.com/bug/6702274).
It is only a doc issue, not functional, if we agree that the code implementation is the source of truth.
Thanks for looking at it.


wsttiger (approved):
LGTM

PR #47, opened 2026-09-08
Add reversible integer arithmetic primitives

wsttiger:
Adds `cudaq_algorithms.primitives` reversible integer arithmetic device kernels — little-endian, in-place, with hand-written inverses (no `cudaq.adjoint`, per cuda-quantum#4897/#4898).

**CDKM/Cuccaro ripple-carry** (arXiv:quant-ph/0410184):
- `add_register` / `subtract_register` — `b ← (a ± b) mod 2ⁿ`, **2n−2** Toffolis (the unused top carry-out cancels; §4.1's 2n−3 not adopted).
- `add_constant` / `subtract_constant` — constant loaded into a caller-provided `work`+`carry` register, 2n−2.
- `cmp_ge_constant` — `out ^= (x ≥ K)`, restores `x`, **2n−1** Toffolis.
- `cmp_ge_register` / `cmp_gt_register` — `out ^= (a ≥ b)` / `(a > b)` on equal-width unsigned registers, 2n−1, no n-qubit work register.

**Draper QFT** (arXiv:quant-ph/0008033, no Toffolis — `r1` rotations only):
- `add_constant_qft` / `subtract_constant_qft`.
- `cmp_ge_constant_qft_shift` / `cmp_ge_constant_qft_shift_adj` — the compute half **leaves `x` shifted** by −K (for the #48 alias sampler); the adjoint restores it.

**Tests (92):** operation-pinning harnesses — fixed-operand superposition for the register ops, basis-input truth tables for the constant ops, correlated-flag superposition for the comparators (over `out ∈ {0,1}` and `K ∈ 0..2ⁿ`) — op-then-inverse identity checks, and `estimate_resources` contracts against the documented per-kernel counts.

**Follow-ups:** ancilla-light constant adder (Häner, arXiv:1611.07995); Cuccaro §4.1's 2n−3 adder; `cudaq.control` composition once the compiler regression is fixed (issue forthcoming).

### DIFF
--- a/python/cudaq_algorithms/primitives/_arithmetic.py
+++ b/python/cudaq_algorithms/primitives/_arithmetic.py
@@ -8,15 +8,22 @@ bit, matching ``docs/conventions.md``):
 - **CDKM/Cuccaro ripple-carry** (`arXiv:quant-ph/0410184`):
   ``add_register`` / ``subtract_register`` (in-place ``b <- b +/- a``),
   ``add_constant`` / ``subtract_constant`` (constant loaded into a caller
-  provided work register), and ``cmp_ge_constant`` (a ``x >= K`` comparator
+  provided work register), ``cmp_ge_constant`` (a ``x >= K`` comparator
   writing into an out qubit, leaving ``x`` untouched: MAJ sweep, copy the
-  carry, reverse MAJ sweep). Toffoli prices (documented contracts, pinned
-  by the resource tests in ``tests/python/test_primitives_arithmetic.py``):
+  carry, reverse MAJ sweep), and the register-register comparators
+  ``cmp_ge_register`` / ``cmp_gt_register`` (``out ^= (a >= b)`` /
+  ``out ^= (a > b)`` on equal-width unsigned registers, same
+  MAJ-copy-reverse structure with ``b`` complemented in place, so no
+  ``n``-qubit work register at all — only the one-qubit carry). Toffoli
+  prices (documented contracts, pinned by the resource tests in
+  ``tests/python/test_primitives_arithmetic.py``):
   ``add_register`` / ``subtract_register`` cost exactly ``2 n`` Toffolis
   on ``n``-bit registers (``n`` in the MAJ sweep, ``n`` in the UMA sweep),
   ``add_constant`` / ``subtract_constant`` inherit the same ``2 n`` (the
-  constant load is X-only), and ``cmp_ge_constant`` costs ``2 n`` for
-  ``K >= 1`` (MAJ sweep plus its reversal) and ``0`` for ``K = 0``.
+  constant load is X-only), ``cmp_ge_constant`` costs ``2 n`` for
+  ``K >= 1`` (MAJ sweep plus its reversal) and ``0`` for ``K = 0``, and
+  ``cmp_ge_register`` / ``cmp_gt_register`` cost exactly ``2 n`` (the
+  complement and flag copy are X/CNOT-only).
 - **Draper QFT arithmetic** (`arXiv:quant-ph/0008033`): ``qft`` / ``iqft``
   and ``add_constant_qft``, plus the ``cmp_ge_constant_qft`` /
   ``cmp_ge_constant_qft_adj`` comparator pair. The QFT family needs *no*
@@ -55,6 +62,8 @@ __all__ = [
     "add_constant",
     "subtract_constant",
     "cmp_ge_constant",
+    "cmp_ge_register",
+    "cmp_gt_register",
     "qft",
     "iqft",
     "phase_add_constant",
@@ -210,6 +219,70 @@ def cmp_ge_constant(x_reg: cudaq.qview, complement_bits: list[int],
         for k in range(n):
             if complement_bits[k] == 1:
                 x(work[k])
+
+
+@cudaq.kernel
+def cmp_ge_register(a: cudaq.qview, b: cudaq.qview, carry: cudaq.qview,
+                    out: cudaq.qview):
+    """``out[0] ^= (a >= b)`` (CDKM, unsigned), leaving ``a``, ``b`` unchanged.
+
+    ``a`` and ``b`` have equal size ``n >= 1``; ``carry`` is a one-qubit
+    view that must be |0> on entry (it is returned to |0>); ``out`` is the
+    one-qubit flag, XOR-loaded (any input state is allowed). ``a``, ``b``,
+    ``carry`` and ``out`` must be pairwise disjoint (module precondition).
+
+    Uses the ripple identity ``a >= b <=> carry_out(a + (2^n - b))`` with
+    ``2^n - b = ~b + 1``: complement ``b`` in place and set the carry-in
+    (so ``b`` doubles as the constant-load work register — no extra
+    ``n``-qubit work is needed, unlike ``cmp_ge_constant``), MAJ sweep,
+    copy the carry-out (on the top ``b`` qubit) into ``out``, reverse the
+    MAJ sweep and undo the complement so ``a``, ``b`` and ``carry`` are
+    all restored. Because the flag is XOR-accumulated and the operand
+    action is compute-copy-uncompute, applying the kernel twice with the
+    same arguments is the identity (self-inverse).
+    """
+    n = a.size()
+    if n > 0:
+        # 2^n - b = ~b + 1: complement b in place, set the carry-in to 1.
+        for k in range(n):
+            x(b[k])
+        x(carry[0])
+        # MAJ sweep of a + ~b + 1 (b accumulates the ripple carries).
+        cx(b[0], a[0])
+        cx(b[0], carry[0])
+        x.ctrl(carry[0], a[0], b[0])
+        for i in range(1, n):
+            cx(b[i], a[i])
+            cx(b[i], b[i - 1])
+            x.ctrl(b[i - 1], a[i], b[i])
+        # a >= b iff the sum carries out of the top bit.
+        cx(b[n - 1], out[0])
+        # Reverse MAJ sweep (restores a, ~b and the carry-in).
+        for k in range(1, n):
+            i = n - k
+            x.ctrl(b[i - 1], a[i], b[i])
+            cx(b[i], b[i - 1])
+            cx(b[i], a[i])
+        x.ctrl(carry[0], a[0], b[0])
+        cx(b[0], carry[0])
+        cx(b[0], a[0])
+        x(carry[0])
+        for k in range(n):
+            x(b[k])
+
+
+@cudaq.kernel
+def cmp_gt_register(a: cudaq.qview, b: cudaq.qview, carry: cudaq.qview,
+                    out: cudaq.qview):
+    """``out[0] ^= (a > b)`` (CDKM, unsigned), leaving ``a``, ``b`` unchanged.
+
+    Free strict variant of ``cmp_ge_register`` via ``a > b <=> not
+    (b >= a)``: an X on ``out`` plus the ``>=`` comparator with the roles
+    swapped. Same preconditions, same ``2 n`` Toffoli price, and likewise
+    self-inverse.
+    """
+    x(out[0])
+    cmp_ge_register(b, a, carry, out)
 
 
 # ============================================================================
--- a/python/cudaq_algorithms/primitives/_arithmetic.py
+++ b/python/cudaq_algorithms/primitives/_arithmetic.py
@@ -1,0 +1,333 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""Reversible in-place integer arithmetic device kernels.
+
+Two families, both little-endian (``register[0]`` is the least-significant
+bit, matching ``docs/conventions.md``):
+
+- **CDKM/Cuccaro ripple-carry** (`arXiv:quant-ph/0410184`):
+  ``add_register`` / ``subtract_register`` (in-place ``b <- b +/- a``),
+  ``add_constant`` / ``subtract_constant`` (constant loaded into a caller
+  provided work register), and ``cmp_ge_constant`` (a ``x >= K`` comparator
+  writing into an out qubit, leaving ``x`` untouched: MAJ sweep, copy the
+  carry, reverse MAJ sweep). Toffoli prices (documented contracts, pinned
+  by the resource tests in ``tests/python/test_primitives_arithmetic.py``):
+  ``add_register`` / ``subtract_register`` cost exactly ``2 n`` Toffolis
+  on ``n``-bit registers (``n`` in the MAJ sweep, ``n`` in the UMA sweep),
+  ``add_constant`` / ``subtract_constant`` inherit the same ``2 n`` (the
+  constant load is X-only), and ``cmp_ge_constant`` costs ``2 n`` for
+  ``K >= 1`` (MAJ sweep plus its reversal) and ``0`` for ``K = 0``.
+- **Draper QFT arithmetic** (`arXiv:quant-ph/0008033`): ``qft`` / ``iqft``
+  and ``add_constant_qft``, plus the ``cmp_ge_constant_qft`` /
+  ``cmp_ge_constant_qft_adj`` comparator pair. The QFT family needs *no*
+  work qubits — on a statevector simulator every extra work qubit doubles
+  the memory, and the CDKM constant ops need ``n + 1`` of them — and *no*
+  Toffolis: its price is rotations, ``n (n - 1)`` controlled-``r1`` plus
+  ``n`` free ``r1`` for ``add_constant_qft`` on ``n`` bits, and
+  ``n (n + 1)`` controlled-``r1`` plus ``n + 1`` free ``r1`` per side of
+  the ``cmp_ge_constant_qft`` pair (``K >= 1``; zero rotations at
+  ``K = 0``), also pinned by the resource tests.
+
+Registers passed to any one op must be pairwise disjoint: the kernels
+never alias-check, and overlapping views (e.g. ``add_register(a, a,
+carry)``, or a ``work`` view sharing qubits with ``target``) are
+undefined.
+
+Every inverse is hand-written (``cudaq.adjoint`` is off-limits as of
+CUDA-Q 0.15: cuda-quantum#4897/#4898, still unresolved as of the 0.16
+pre-release) and pinned by op-then-inverse identity tests in
+``tests/python/test_primitives_arithmetic.py``, alongside exhaustive
+value checks against classical integer arithmetic for widths up to 5.
+
+Kernel-language notes (see CLAUDE.md): guards are positive ``if`` blocks
+(kernel ``return`` is silently ignored, cuda-quantum#4845), and all
+classical indices are recomputed from loop variables rather than carried
+as mutating accumulators.
+"""
+
+from __future__ import annotations
+
+import cudaq
+
+__all__ = [
+    "add_register",
+    "subtract_register",
+    "add_constant",
+    "subtract_constant",
+    "cmp_ge_constant",
+    "qft",
+    "iqft",
+    "phase_add_constant",
+    "add_constant_qft",
+    "subtract_constant_qft",
+    "cmp_ge_constant_qft",
+    "cmp_ge_constant_qft_adj",
+]
+
+# ============================================================================
+# CDKM / Cuccaro ripple-carry family
+# ============================================================================
+#
+# MAJ(x, y, z) = cx(z, y); cx(z, x); ccx(x, y, z)
+# UMA(x, y, z) = ccx(x, y, z); cx(z, x); cx(x, y)
+#
+# The adder chains MAJ(carry, b0, a0), MAJ(a0, b1, a1), ...; after the MAJ
+# sweep the ripple carry-out sits on a[n-1]; the UMA sweep (in reverse)
+# restores ``a`` and the carry ancilla and completes ``b <- a + b mod 2^n``.
+
+
+@cudaq.kernel
+def add_register(a: cudaq.qview, b: cudaq.qview, carry: cudaq.qview):
+    """``b <- (a + b) mod 2^n`` (CDKM). ``a`` and ``carry`` are restored.
+
+    ``a`` and ``b`` have equal size ``n``; ``carry`` is a one-qubit view
+    that must be |0> on entry (it is returned to |0>).
+    """
+    n = a.size()
+    if n > 0:
+        # MAJ sweep.
+        cx(a[0], b[0])
+        cx(a[0], carry[0])
+        x.ctrl(carry[0], b[0], a[0])
+        for i in range(1, n):
+            cx(a[i], b[i])
+            cx(a[i], a[i - 1])
+            x.ctrl(a[i - 1], b[i], a[i])
+        # UMA sweep (reverse).
+        for k in range(1, n):
+            i = n - k
+            x.ctrl(a[i - 1], b[i], a[i])
+            cx(a[i], a[i - 1])
+            cx(a[i - 1], b[i])
+        x.ctrl(carry[0], b[0], a[0])
+        cx(a[0], carry[0])
+        cx(carry[0], b[0])
+
+
+@cudaq.kernel
+def subtract_register(a: cudaq.qview, b: cudaq.qview, carry: cudaq.qview):
+    """``b <- (b - a) mod 2^n``: the exact gate-reversal of ``add_register``.
+
+    Hand-written inverse (no ``cudaq.adjoint``); every gate of the adder is
+    self-inverse, so the reversed sequence is the inverse circuit.
+    """
+    n = a.size()
+    if n > 0:
+        cx(carry[0], b[0])
+        cx(a[0], carry[0])
+        x.ctrl(carry[0], b[0], a[0])
+        for i in range(1, n):
+            cx(a[i - 1], b[i])
+            cx(a[i], a[i - 1])
+            x.ctrl(a[i - 1], b[i], a[i])
+        for k in range(1, n):
+            i = n - k
+            x.ctrl(a[i - 1], b[i], a[i])
+            cx(a[i], a[i - 1])
+            cx(a[i], b[i])
+        x.ctrl(carry[0], b[0], a[0])
+        cx(a[0], carry[0])
+        cx(a[0], b[0])
+
+
+@cudaq.kernel
+def add_constant(target: cudaq.qview, constant_bits: list[int],
+                 work: cudaq.qview, carry: cudaq.qview):
+    """``target <- (target + K) mod 2^n`` (CDKM).
+
+    ``constant_bits[k]`` is bit ``k`` of ``K`` (little-endian). Precondition:
+    ``constant_bits`` has length exactly ``n = target.size()``, i.e.
+    ``0 <= K < 2^n`` — a shorter or longer list is not truncated or padded
+    and the kernel is undefined. ``work`` (``n`` qubits) and ``carry`` (1
+    qubit) must be |0> on entry and are returned to |0>: the constant is
+    X-loaded into ``work``, ripple-added, and X-unloaded.
+    """
+    n = target.size()
+    for k in range(n):
+        if constant_bits[k] == 1:
+            x(work[k])
+    add_register(work, target, carry)
+    for k in range(n):
+        if constant_bits[k] == 1:
+            x(work[k])
+
+
+@cudaq.kernel
+def subtract_constant(target: cudaq.qview, constant_bits: list[int],
+                      work: cudaq.qview, carry: cudaq.qview):
+    """``target <- (target - K) mod 2^n``: inverse of ``add_constant``."""
+    n = target.size()
+    for k in range(n):
+        if constant_bits[k] == 1:
+            x(work[k])
+    subtract_register(work, target, carry)
+    for k in range(n):
+        if constant_bits[k] == 1:
+            x(work[k])
+
+
+@cudaq.kernel
+def cmp_ge_constant(x_reg: cudaq.qview, complement_bits: list[int],
+                    k_is_zero: int, work: cudaq.qview, carry: cudaq.qview,
+                    out: cudaq.qview):
+    """``out[0] ^= (x >= K)`` (CDKM), leaving ``x_reg`` unchanged.
+
+    ``complement_bits`` are the little-endian bits of ``2^n - K``.
+    Precondition: ``complement_bits`` has length exactly ``n =
+    x_reg.size()`` and ``0 <= K <= 2^n`` (so ``2^n - K`` fits in ``n``
+    bits; pass ``k_is_zero = 1`` and all-zero bits for ``K = 0``, which is
+    always true, and all-zero bits with ``k_is_zero = 0`` for ``K = 2^n``,
+    which is always false). Uses the ripple identity ``x >= K <=> carry_out(x + (2^n
+    - K))`` for ``K >= 1``: MAJ sweep, copy the carry-out (which sits on
+    the top work qubit) into ``out``, then reverse the MAJ sweep so
+    ``x_reg``, ``work`` and ``carry`` are all restored.
+    """
+    if k_is_zero == 1:
+        x(out[0])
+    else:
+        n = x_reg.size()
+        for k in range(n):
+            if complement_bits[k] == 1:
+                x(work[k])
+        # MAJ sweep (a = work, b = x_reg).
+        cx(work[0], x_reg[0])
+        cx(work[0], carry[0])
+        x.ctrl(carry[0], x_reg[0], work[0])
+        for i in range(1, n):
+            cx(work[i], x_reg[i])
+            cx(work[i], work[i - 1])
+            x.ctrl(work[i - 1], x_reg[i], work[i])
+        cx(work[n - 1], out[0])
+        # Reverse MAJ sweep (restores x_reg, work, carry).
+        for k in range(1, n):
+            i = n - k
+            x.ctrl(work[i - 1], x_reg[i], work[i])
+            cx(work[i], work[i - 1])
+            cx(work[i], x_reg[i])
+        x.ctrl(carry[0], x_reg[0], work[0])
+        cx(work[0], carry[0])
+        cx(work[0], x_reg[0])
+        for k in range(n):
+            if complement_bits[k] == 1:
+                x(work[k])
+
+
+# ============================================================================
+# Draper QFT family (no work qubits)
+# ============================================================================
+#
+# ``qft`` is the textbook circuit without the final bit-reversal swaps;
+# after it, qubit t holds (|0> + exp(2 pi i x / 2^(t+1)) |1>)/sqrt(2), so a
+# constant K is added by the single-qubit phases r1(2 pi K / 2^(t+1)).
+
+
+@cudaq.kernel
+def qft(reg: cudaq.qview):
+    """Quantum Fourier transform (no bit-reversal swaps; see module doc)."""
+    n = reg.size()
+    for j in range(n):
+        t = n - 1 - j
+        h(reg[t])
+        for c in range(t):
+            r1.ctrl(3.141592653589793 / (1 << (t - c)), reg[c], reg[t])
+
+
+@cudaq.kernel
+def iqft(reg: cudaq.qview):
+    """Inverse QFT: hand-written reversal of ``qft``."""
+    n = reg.size()
+    for t in range(n):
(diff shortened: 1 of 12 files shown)


kvmto (commented):
Questions on the PR description:

- `add_constant` and `cmp_ge_constant` are listed without `work` and `carry`. Which signature is right?
- `cmp_ge_register` and `cmp_gt_register` are missing from the list. Can you add them?
- It says 62 tests. The file collects 96. Which is it?
- "Exhaustive truth tables for every operation." Which adder test is a truth table?
- "Random superposed states" in the adjointness checks. Only the register comparators use random states. Can we reword?
- The design note says these compose under `cudaq.control`. On 0.15.1 they do not. Can we remove or qualify this until there is a test?

Inline questions on the code and tests below.

python/cudaq_algorithms/primitives/_arithmetic.py
```diff
@@ -0,0 +1,406 @@
...
+CUDA-Q 0.15: cuda-quantum#4897/#4898, still unresolved as of the 0.16
+pre-release) and pinned by op-then-inverse identity tests in
+``tests/python/test_primitives_arithmetic.py``, alongside exhaustive
+value checks against classical integer arithmetic for widths up to 5.
+
+Kernel-language notes (see CLAUDE.md): guards are positive ``if`` blocks
+(kernel ``return`` is silently ignored, cuda-quantum#4845), and all
+classical indices are recomputed from loop variables rather than carried
+as mutating accumulators.
+"""
```
kvmto, line 17: Can we cut this docstring to a few lines plus the two arXiv links, and move each gate price to the kernel that owns it? Same house style as PR #40.

python/cudaq_algorithms/primitives/_arithmetic.py
```diff
@@ -0,0 +1,406 @@
...
+
+Registers passed to any one op must be pairwise disjoint: the kernels
+never alias-check, and overlapping views (e.g. ``add_register(a, a,
+carry)``, or a ``work`` view sharing qubits with ``target``) are
+undefined.
+
+Every inverse is hand-written (``cudaq.adjoint`` is off-limits as of
+CUDA-Q 0.15: cuda-quantum#4897/#4898, still unresolved as of the 0.16
+pre-release) and pinned by op-then-inverse identity tests in
+``tests/python/test_primitives_arithmetic.py``, alongside exhaustive
```
kvmto: We distrust `cudaq.adjoint` here. Why do we trust `cudaq.control` without a test?

python/cudaq_algorithms/primitives/_arithmetic.py
```diff
@@ -0,0 +1,406 @@
...
+        cx(a[0], carry[0])
+        x.ctrl(carry[0], b[0], a[0])
+        for i in range(1, n):
+            cx(a[i], b[i])
+            cx(a[i], a[i - 1])
+            x.ctrl(a[i - 1], b[i], a[i])
+        # UMA sweep (reverse).
+        for k in range(1, n):
+            i = n - k
+            x.ctrl(a[i - 1], b[i], a[i])
```
kvmto, line 86: This Toffoli and the one at L104 are the same gate on the same qubits with nothing in between. Don't they cancel?

After removing them, the two `cx(a[i], a[i-1])` at the top bit also cancel. Is the count then 2n-2, not 2n?

python/cudaq_algorithms/primitives/_arithmetic.py
```diff
@@ -0,0 +1,406 @@
...
+# MAJ(x, y, z) = cx(z, y); cx(z, x); ccx(x, y, z)
+# UMA(x, y, z) = ccx(x, y, z); cx(z, x); cx(x, y)
+#
+# The adder chains MAJ(carry, b0, a0), MAJ(a0, b1, a1), ...; after the MAJ
+# sweep the ripple carry-out sits on a[n-1]; the UMA sweep (in reverse)
+# restores ``a`` and the carry ancilla and completes ``b <- a + b mod 2^n``.
+
+
+@cudaq.kernel
+def add_register(a: cudaq.qview, b: cudaq.qview, carry: cudaq.qview):
```
kvmto, line 53: For n = 1 this emits two Toffolis. Isn't a 1-bit mod-2 adder a single CNOT?

python/cudaq_algorithms/primitives/_arithmetic.py
```diff
@@ -0,0 +1,406 @@
...
+  writing into an out qubit, leaving ``x`` untouched: MAJ sweep, copy the
+  carry, reverse MAJ sweep), and the register-register comparators
+  ``cmp_ge_register`` / ``cmp_gt_register`` (``out ^= (a >= b)`` /
+  ``out ^= (a > b)`` on equal-width unsigned registers, same
+  MAJ-copy-reverse structure with ``b`` complemented in place, so no
+  ``n``-qubit work register at all — only the one-qubit carry). Toffoli
+  prices (documented contracts, pinned by the resource tests in
+  ``tests/python/test_primitives_arithmetic.py``):
+  ``add_register`` / ``subtract_register`` cost exactly ``2 n`` Toffolis
+  on ``n``-bit registers (``n`` in the MAJ sweep, ``n`` in the UMA sweep),
```
kvmto: Cuccaro Table 1 gives 2n-3 Toffolis for a mod 2^n adder. Why do we pin 2n as the contract?

python/cudaq_algorithms/primitives/_arithmetic.py
```diff
@@ -0,0 +1,406 @@
...
+        cx(carry[0], b[0])
+        cx(a[0], carry[0])
+        x.ctrl(carry[0], b[0], a[0])
+        for i in range(1, n):
+            cx(a[i - 1], b[i])
+            cx(a[i], a[i - 1])
+            x.ctrl(a[i - 1], b[i], a[i])
+        for k in range(1, n):
+            i = n - k
+            x.ctrl(a[i - 1], b[i], a[i])
```
kvmto, line 117: Same redundant pair as L104/L108 in `subtract_register`?

python/cudaq_algorithms/primitives/_arithmetic.py
```diff
@@ -0,0 +1,406 @@
...
+            cx(a[i], a[i - 1])
+            cx(a[i], b[i])
+        x.ctrl(carry[0], b[0], a[0])
+        cx(a[0], carry[0])
+        cx(a[0], b[0])
+
+
+@cudaq.kernel
+def add_constant(target: cudaq.qview, constant_bits: list[int],
```
kvmto, line 126: The constant adder loads K into n work qubits. Is this a placeholder? Should we cite an ancilla-light constant adder like Häner et al. (arXiv:1611.07995) as the follow-up?

Also: `constant_bits` is a runtime list. Does the compiler get to simplify anything when K is classical?

python/cudaq_algorithms/primitives/_arithmetic.py
```diff
@@ -0,0 +1,406 @@
...
+
+    ``complement_bits`` are the little-endian bits of ``2^n - K``.
+    Precondition: ``complement_bits`` has length exactly ``n =
+    x_reg.size()`` and ``0 <= K <= 2^n`` (so ``2^n - K`` fits in ``n``
+    bits; pass ``k_is_zero = 1`` and all-zero bits for ``K = 0``, which is
+    always true, and all-zero bits with ``k_is_zero = 0`` for ``K = 2^n``,
+    which is always false). Uses the ripple identity ``x >= K <=> carry_out(x + (2^n
+    - K))`` for ``K >= 1``: MAJ sweep, copy the carry-out (which sits on
+    the top work qubit) into ``out``, then reverse the MAJ sweep so
+    ``x_reg``, ``work`` and ``carry`` are all restored.
```
kvmto: Where does it say `work` and `carry` must be |0> on entry? Restored is not the same as zero-initialised.

python/cudaq_algorithms/primitives/_arithmetic.py
```diff
@@ -0,0 +1,406 @@
...
+        cx(work[0], x_reg[0])
+        cx(work[0], carry[0])
+        x.ctrl(carry[0], x_reg[0], work[0])
+        for i in range(1, n):
+            cx(work[i], x_reg[i])
+            cx(work[i], work[i - 1])
+            x.ctrl(work[i - 1], x_reg[i], work[i])
+        cx(work[n - 1], out[0])
+        # Reverse MAJ sweep (restores x_reg, work, carry).
+        for k in range(1, n):
```
kvmto: CCX, CX to `out`, CCX. Isn't this the same as CX to `out` then CCX into `out` directly? That saves one Toffoli.

python/cudaq_algorithms/primitives/_arithmetic.py
```diff
@@ -0,0 +1,406 @@
...
+        if constant_bits[k] == 1:
+            x(work[k])
+    subtract_register(work, target, carry)
+    for k in range(n):
+        if constant_bits[k] == 1:
+            x(work[k])
+
+
+@cudaq.kernel
+def cmp_ge_constant(x_reg: cudaq.qview, complement_bits: list[int],
```
kvmto, line 169: Two comparators, two contracts (see `cmp_ge_constant_qft` at L365). One XORs the flag, one overwrites it. One leaves x alone, one shifts it. One takes complement bits, one takes an int. Is this intentional? Can we state it in one place?

python/cudaq_algorithms/primitives/_arithmetic.py
```diff
@@ -0,0 +1,406 @@
...
+        # MAJ sweep of a + ~b + 1 (b accumulates the ripple carries).
+        cx(b[0], a[0])
+        cx(b[0], carry[0])
+        x.ctrl(carry[0], a[0], b[0])
+        for i in range(1, n):
+            cx(b[i], a[i])
+            cx(b[i], b[i - 1])
+            x.ctrl(b[i - 1], a[i], b[i])
+        # a >= b iff the sum carries out of the top bit.
+        cx(b[n - 1], out[0])
```
kvmto, line 282: Same one-Toffoli saving as L209?

python/cudaq_algorithms/primitives/_arithmetic.py
```diff
@@ -0,0 +1,406 @@
...
+def cmp_gt_register(a: cudaq.qview, b: cudaq.qview, carry: cudaq.qview,
+                    out: cudaq.qview):
+    """``out[0] ^= (a > b)`` (CDKM, unsigned), leaving ``a``, ``b`` unchanged.
+
+    Free strict variant of ``cmp_ge_register`` via ``a > b <=> not
+    (b >= a)``: an X on ``out`` plus the ``>=`` comparator with the roles
+    swapped. Same preconditions, same ``2 n`` Toffoli price, and likewise
+    self-inverse.
+    """
+    x(out[0])
```
kvmto, line 310: Why is this a separate kernel? It is one X plus a swap of arguments. Could a note on `cmp_ge_register` do?

python/cudaq_algorithms/primitives/_arithmetic.py
```diff
@@ -0,0 +1,406 @@
...
+            r1.ctrl(-3.141592653589793 / (1 << (t - c)), reg[c], reg[t])
+        h(reg[t])
+
+
+@cudaq.kernel
+def phase_add_constant(reg: cudaq.qview, constant: int):
+    """``reg <- reg + K mod 2^n`` in the Fourier basis (between qft/iqft)."""
+    n = reg.size()
+    for t in range(n):
+        r1(6.283185307179586 * constant / (1 << (t + 1)), reg[t])
```
kvmto: K goes to float before the rotation. What happens for K above about 2^40? Can we reduce K mod 2^n as an int first?

python/cudaq_algorithms/primitives/_arithmetic.py
```diff
@@ -0,0 +1,406 @@
...
+
+    Precondition: ``0 <= K <= 2^n`` (``n = x_reg.size()``). For ``K >
+    2^n`` the result is silently wrong: the subtraction acts on the
+    (n+1)-bit extension, so the borrow wraps mod ``2^(n+1)`` and ``out``
+    no longer encodes ``x < K``.
+
+    Draper-style: subtract ``K`` on the (n+1)-bit register ``[x_reg,
+    out]``; the MSB (``out``) becomes the borrow, i.e. ``x < K``. ``out``
+    must be |0> on entry. Between this kernel and
+    ``cmp_ge_constant_qft_adj`` the low bits hold ``(x - K) mod 2^n`` —
```
kvmto: `out` must be |0> here but is XOR-loaded in the CDKM version. Why the difference?

python/cudaq_algorithms/primitives/_arithmetic.py
```diff
@@ -0,0 +1,406 @@
...
+    callers may read ``out`` but must not consume ``x_reg`` until the
+    adjoint restores it. ``K = 0`` (with ``invert = 0``) yields the
+    constant-true comparator.
```
kvmto: The caller must not read x between compute and adjoint. Which consumer needs the shift? Is it the alias sampler in PR #48?

Can we call this something other than `cmp_ge` to make clear it changes x?

tests/python/test_primitives_arithmetic.py
```diff
@@ -0,0 +1,577 @@
...
+# SPDX-License-Identifier: Apache-2.0
+"""Exhaustive tests for the primitives arithmetic device kernels.
+
+Every operation is checked against classical integer arithmetic over *all*
+inputs for register widths up to 5, and every inverse is pinned by an
+op-then-inverse == identity test. Exhaustiveness comes from a superposition
+harness: the input registers are prepared in a uniform superposition, so a
+single statevector comparison validates the operation's action on every
+computational-basis input at once (the operations are classical
+permutations, so distinct inputs cannot interfere).
```
kvmto: A uniform superposition is the same after any permutation. What does this test prove about addition?

tests/python/test_primitives_arithmetic.py
```diff
@@ -0,0 +1,577 @@
...
+
+def _register_op_expected(n: int, op) -> np.ndarray:
+    """Uniform superposition over (a, b) mapped through (a, op(a, b))."""
+    expected = np.zeros(1 << (2 * n + 1), dtype=np.complex128)
+    norm = 1.0 / (1 << n)
+    for a in range(1 << n):
+        for b in range(1 << n):
+            expected[a + (op(a, b) % (1 << n)) * (1 << n)] += norm
+    return expected
+
```
kvmto: This expected vector is the same for add, subtract and identity. Is that intended?

tests/python/test_primitives_arithmetic.py
```diff
@@ -0,0 +1,577 @@
...
+@pytest.mark.parametrize("n", WIDTHS)
+def test_subtract_register_all_inputs(n):
+    state = np.array(cudaq.get_state(_run_add_register, n, 1))
+    np.testing.assert_allclose(state,
+                               _register_op_expected(n, lambda a, b: b - a),
+                               atol=1e-12)
+
+
+@pytest.mark.parametrize("n", WIDTHS)
+def test_subtract_register_inverts_add_register(n):
```
kvmto, line 112: I swapped `subtract_register` for `add_register` and this, L80 and L88 still passed. What is the test actually checking?

tests/python/test_primitives_arithmetic.py
```diff
@@ -0,0 +1,577 @@
...
+            arith.add_constant_qft(target, constant)
+
+
+def _constant_op_expected(n: int, shift: int, extra_qubits: int) -> np.ndarray:
+    expected = np.zeros(1 << (n + extra_qubits), dtype=np.complex128)
+    norm = 1.0 / np.sqrt(1 << n)
+    for value in range(1 << n):
+        expected[(value + shift) % (1 << n)] += norm
+    return expected
+
```
kvmto: Does the `shift` argument change the output at all?

tests/python/test_primitives_arithmetic.py
```diff
@@ -0,0 +1,577 @@
...
+
+def _bits(value: int, n: int) -> list[int]:
+    return [(value >> k) & 1 for k in range(n)]
+
+
+@pytest.mark.parametrize("n", WIDTHS)
+def test_add_and_subtract_constant_all_inputs(n):
```
kvmto, line 178: Same question as the register tests for the constant adders (also L178). Can we run a truth table like L346 does?

tests/python/test_primitives_arithmetic.py
```diff
@@ -0,0 +1,577 @@
...
+    for constant in range(1 << n):
```
kvmto, line 179: The loop stops at K = 2^n - 1 (also L179, L238, L254). The docstrings promise K = 2^n. Where is the boundary tested?

tests/python/test_primitives_arithmetic.py
```diff
@@ -0,0 +1,577 @@
...
+        complement = _bits((1 << n) - constant, n) if constant else [0] * n
+        state = np.array(
+            cudaq.get_state(_run_cmp_ge_constant, n, complement,
+                            int(constant == 0)))
+        np.testing.assert_allclose(state,
+                                   _cmp_expected(n, lambda v: v >= constant,
+                                                 n + 1),
+                                   atol=1e-12)
+
```
kvmto, line 277: `out` is only tested at |0>. The docstring says `out ^= result`. Where is `out = 1` tested?

tests/python/test_primitives_arithmetic.py
```diff
@@ -0,0 +1,577 @@
...
+_CMP_REGISTER_OPS = [
+    ("ge", 0, lambda a, b: a >= b),
+    ("gt", 1, lambda a, b: a > b),
+]
+
+
+@pytest.mark.parametrize("n", [1, 2, 3, 4])
+@pytest.mark.parametrize("name,strict,predicate", _CMP_REGISTER_OPS)
+def test_cmp_register_truth_table_exhaustive(n, name, strict, predicate):
```
kvmto, line 378: These are the strongest tests in the file. Can the adder tests follow this pattern?

tests/python/test_primitives_arithmetic.py
```diff
@@ -0,0 +1,577 @@
...
+        if ((bval >> k) & 1) == 1:
+            x(b[k])
+    arith.add_register(a, b, carry)
+
+
+def test_add_register_basis_spot_checks():
+    for n, aval, bval in [(2, 1, 2), (3, 5, 6), (5, 21, 27)]:
+        state = np.array(cudaq.get_state(_spot_add, n, aval, bval))
+        index = aval + (((aval + bval) % (1 << n)) << n)
+        np.testing.assert_allclose(state, _basis(index, 2 * n + 1), atol=1e-12)
```
kvmto, line 446: Only three spot checks and only for `add_register`. Why not a full truth table at small n?

tests/python/test_primitives_arithmetic.py
```diff
@@ -0,0 +1,577 @@
...
+def _toffolis(kernel, *args) -> int:
+    # A Toffoli is an x with two controls; ``count_controls`` is the
+    # arity-aware accessor (``count("ccx")`` matches nothing — the
+    # display name is not the lookup key and returns 0).
+    return cudaq.estimate_resources(kernel, *args).count_controls("x", 2)
+
+
+@_RESOURCES
+@pytest.mark.parametrize("n", WIDTHS)
+def test_cdkm_register_ops_cost_exactly_2n_toffolis(n):
```
kvmto: If the adder becomes 2n-2 or 2n-3, these all change. Should the contract reference Cuccaro Table 1 rather than a derived number?

tests/python/test_primitives_arithmetic.py
```diff
@@ -0,0 +1,577 @@
...
+
+
+@_RESOURCES
+def test_cdkm_toffoli_cost_grows_linearly_per_doubling():
+    # The cost is exactly linear (2 n): each width doubling doubles it.
+    compiled = {n: _toffolis(_res_add_register, n, 0) for n in (2, 4, 8, 16)}
+    for n in (2, 4, 8):
+        assert compiled[2 * n] == 2 * compiled[n], compiled
+
+
```
kvmto: The doubling law only holds for exactly 2n. Is this test asserting the optimisation is absent?

tests/python/test_primitives_arithmetic.py
```diff
@@ -0,0 +1,577 @@
...
+"""
+
+import numpy as np
+import pytest
+
+import cudaq
+
+from cudaq_algorithms.primitives import _arithmetic as arith
+
+WIDTHS = [1, 2, 3, 4, 5]
```
kvmto, line 34: Where is the `cudaq.control` test? On 0.15.1, direct control of seven of these kernels fails to compile, and control through a one-line wrapper compiles but fires on both branches. Have you seen this? Should we file it upstream?

Sam asked in PR #40 for tests with a superposed control. Do we have one here?

kvmto (commented):
Questions on the PR description:

- `add_constant` and `cmp_ge_constant` are listed without `work` and `carry`. Which signature is right? The code has both, so the description looks stale.
- `cmp_ge_register` and `cmp_gt_register` are missing from the list. Can you add them?
- It says 62 tests. The file collects 96. Which is it?
- "Exhaustive truth tables for every operation." Which adder test is a truth table? Only the register comparators and the three spot checks are.
- "Random superposed states" in the adjointness checks. Only the register comparators use random states. Can we reword?
- The design note says these compose under `cudaq.control`. On 0.15.1 they do not. Can we remove or qualify this until there is a test?

Inline questions on the code and tests below.

python/cudaq_algorithms/primitives/_arithmetic.py
kvmto, line 17: Can we cut this docstring to a few lines plus the two arXiv links, and move each gate price to the kernel that owns it? Same house style as PR #40. Keep the little-endian convention, the disjointness rule, and the cuda-quantum issue link for hand-written inverses. Everything else belongs on the kernel or in the paper.

python/cudaq_algorithms/primitives/_arithmetic.py
kvmto: We distrust `cudaq.adjoint` here. Why do we trust `cudaq.control` without a test? On CUDA-Q 0.15.1, direct control of `add_constant`, `subtract_constant`, `cmp_gt_register` and the four QFT wrappers fails with "Could not successfully apply kernel specialization". Controlling a one-line wrapper kernel compiles but applies the op on both control branches. 0.14.2 is correct. Worth a numerical test with a superposed control, and an upstream issue.

python/cudaq_algorithms/primitives/_arithmetic.py
kvmto, line 86: This Toffoli and the one at L104 are the same gate on the same qubits with nothing in between. Don't they cancel? At i = n-1 the MAJ Toffoli computes the carry-out c_n into `a[n-1]` and the first UMA Toffoli uncomputes it. A mod 2^n adder never reads c_n. After removing them, the two `cx(a[n-1], a[n-2])` also cancel. What survives at the top bit is `cx(a[n-1], b[n-1])` then `cx(a[n-2], b[n-1])`. I checked this agrees on all basis inputs for n = 2, 3, 4 with 2n-2 Toffolis.

python/cudaq_algorithms/primitives/_arithmetic.py
kvmto, line 53: For n = 1 this emits two Toffolis. Isn't a 1-bit mod-2 adder a single CNOT? Same cancellation as at L104/L108, degenerate case.

python/cudaq_algorithms/primitives/_arithmetic.py
kvmto: Cuccaro Table 1 gives 2n-3 Toffolis for a mod 2^n adder (n >= 3). Why do we pin 2n as the contract? Section 4.1 of the paper runs an (n-1)-bit adder and finishes with one CNOT from `a[n-1]` to `b[n-1]`. Either adopt it, or pin 2n-2 from the cancellation and say why we stop there.

python/cudaq_algorithms/primitives/_arithmetic.py
kvmto, line 117: Same redundant Toffoli pair as L104/L108 in `subtract_register`? It is the exact reversal of `add_register`, so the same fix applies mirrored.

python/cudaq_algorithms/primitives/_arithmetic.py
```diff
@@ -0,0 +1,406 @@
...
+                      work: cudaq.qview, carry: cudaq.qview):
+    """``target <- (target - K) mod 2^n``: inverse of ``add_constant``."""
+    n = target.size()
+    for k in range(n):
+        if constant_bits[k] == 1:
+            x(work[k])
+    subtract_register(work, target, carry)
+    for k in range(n):
+        if constant_bits[k] == 1:
+            x(work[k])
```
kvmto, line 165: The constant adder loads K into n work qubits. Is this a placeholder? There are constant adders in the literature with far fewer ancillas. A sentence saying this is the simple version and what would replace it is enough for now.

Also: `constant_bits` is a runtime list. Does the compiler get to simplify anything when K is classical? With K baked in at kernel construction, MAJ blocks on zero bits collapse. With a runtime list they cannot. Is the runtime list a deliberate choice?

python/cudaq_algorithms/primitives/_arithmetic.py
kvmto: Where does it say `work` and `carry` must be |0> on entry? Restored is not the same as zero-initialised. With work = 0, carry = 1, x = 0, K = 1 the kernel reports 0 >= 1. The wording from `add_register` at L92-93 would do.

python/cudaq_algorithms/primitives/_arithmetic.py
kvmto: CCX, CX to `out`, CCX. Isn't this the same as CX to `out` then CCX into `out` directly? CCX(u,v;w) CX(w;f) CCX(u,v;w) = CX(w;f) CCX(u,v;f), exact on all 16 inputs. Saves one Toffoli, giving 2n-1, which is Cuccaro's comparator count.

python/cudaq_algorithms/primitives/_arithmetic.py
kvmto, line 169: Two comparators, two contracts (this one and `cmp_ge_constant_qft` at L365). One XORs the flag, one overwrites it. One leaves x alone, one shifts it. One takes complement bits plus `k_is_zero`, one takes an int. Is this intentional? A caller switching families to save ancillas must change three things at once. Either unify or put a short comparison table at the top of the module.

python/cudaq_algorithms/primitives/_arithmetic.py
kvmto, line 282: Same one-Toffoli saving as L209?

python/cudaq_algorithms/primitives/_arithmetic.py
kvmto, line 310: Why is `cmp_gt_register` a separate kernel? It is one X plus swapped arguments. A note on `cmp_ge_register` saying "a > b is X(out) then cmp_ge_register(b, a)" may be enough, unless the resource tests want a named entry.

python/cudaq_algorithms/primitives/_arithmetic.py
kvmto: K goes to float before the rotation. What happens for K above about 2^40? 2^40+1 on a 2-qubit zero register gives 1 with only 75% probability, and 2^60+1 gives the wrong integer at 98%. Reducing K mod 2^n as an integer before the multiply fixes it. Subtract passes -K, so mind the sign of the modulo.

python/cudaq_algorithms/primitives/_arithmetic.py
kvmto: `out` must be |0> here but is XOR-loaded in the CDKM version. Why the difference? If the QFT version cannot XOR cheaply, say so in the docstring and in the module-level comparison.

python/cudaq_algorithms/primitives/_arithmetic.py
kvmto: The caller must not read x between compute and adjoint. Which consumer needs the shift? If it is the alias sampler in PR #48, name it. Also consider a name that signals the shift, since `cmp_ge` suggests a clean predicate oracle.

tests/python/test_primitives_arithmetic.py
kvmto: A uniform superposition is the same after any permutation. What does this test prove about addition? It proves the kernel is a phase-free permutation that restores its ancillas. It does not prove which permutation. The docstring claim of exhaustiveness should be reworded or the tests changed.

tests/python/test_primitives_arithmetic.py
kvmto: This expected vector is the same for add, subtract and identity. Is that intended? `op(a, b)` is a bijection for fixed a, so the sum over b is uniform regardless of op.

tests/python/test_primitives_arithmetic.py
kvmto, line 112: I replaced `subtract_register` with a copy of `add_register` and flipped the sign in `subtract_constant_qft`. All 46 adder and spot-check tests (this one, L80, L88 and the constant tests) still passed. What is the test actually checking? The truth-table pattern from L346 at n = 1..4 would catch this, or a reference copy of the input register so the input-output correlation survives.

tests/python/test_primitives_arithmetic.py
kvmto: Does the `shift` argument change the output at all? It does not, for the same reason as L69-77.

tests/python/test_primitives_arithmetic.py
kvmto, line 178: Same question as the register tests for the constant adders (also L178). Can we run a truth table like L346 does? For CDKM at n <= 4 that is at most 16 x 16 states per width, cheap.

tests/python/test_primitives_arithmetic.py
kvmto, line 179: The loop stops at K = 2^n - 1 (also L179, L238, L254). The docstrings promise K = 2^n. Where is the boundary tested? I checked K = 2^n by hand for both comparators at n = 1..3 and it is correct, so this is a coverage gap, not a bug.

tests/python/test_primitives_arithmetic.py
kvmto, line 277: `out` is only tested at |0>. The docstring says `out ^= result`. Where is `out = 1` tested? Also correct by hand, just untested. The `flag_init` loop at L349 is the pattern.

tests/python/test_primitives_arithmetic.py
kvmto, line 378: These are the strongest tests in the file. Can the adder tests follow this pattern?

tests/python/test_primitives_arithmetic.py
kvmto, line 446: Only three spot checks and only for `add_register`. Why not a full truth table at small n? This is the only test that currently distinguishes add from subtract.

tests/python/test_primitives_arithmetic.py
kvmto: If the adder becomes 2n-2 or 2n-3, these all change. Should the contract reference Cuccaro Table 1 rather than a derived number? State the count once per kernel docstring and have the test read that number.

tests/python/test_primitives_arithmetic.py
kvmto: The doubling law only holds for exactly 2n. Is this test asserting the optimisation is absent? 2(2n) - 2 != 2(2n - 2). A direct per-width assertion would be more robust.

tests/python/test_primitives_arithmetic.py
kvmto, line 34: Where is the `cudaq.control` test? Sam asked in PR #40 for a superposed control against an analytic reference. That test would have caught the 0.15.1 control drop described in the `_arithmetic.py` comment at L43-46.

wsttiger:
Thanks @kvmto — this was a genuinely useful review; it caught real problems, especially the tests. Changes pushed in 6a93217; point-by-point below (grouped, since several threads are the same issue).

### PR description
Rewritten to match the code. The stale signatures (`add_constant`/`cmp_ge_constant` do take `work`+`carry`), the missing `cmp_ge_register`/`cmp_gt_register`, and the "62 tests" (it collects 92 now) are all fixed. The "exhaustive truth tables for every operation" and "random superposed states" overclaims are gone.

### Tests — the vacuity (L11/L77/L108, and L174 for the constant ops)
You're right, and thank you: the collapsed uniform-superposition check is invariant under the operation — any bijection sends a uniform input to the same uniform output, so add, subtract and identity all passed. Fixed:
- register add/subtract now **fix `b` and superpose `a`** (looping all `b`): for a fixed `b`, `a→a+b` and `a→b−a` are different permutations, so the statevector distinguishes them, and it fails if the ops are swapped.
- constant ops now use a **basis-input truth table** (a single register has no operand to hold fixed, so a definite input is needed to see `K`).
- comparators: added `out ∈ {0,1}` (the XOR contract, L273) and `K` swept over `0 .. 2^n` inclusive — both the `K=0` and `K=2^n` boundaries (L175).

### Toffoli counts (L83 / L211 / L279 / L513 / L551)
- **Adders**: applied the cancellation you described — the mod-2ⁿ adder never reads the top carry-out, so its MAJ/UMA Toffoli pair and the surrounding CX pair cancel, leaving `cx(a[n-1],b[n-1]); cx(a[n-2],b[n-1])`. `add_register`/`subtract_register` (and the constant adders) are now **2n−2** (n=1→0), verified inverse-preserving by the strengthened tests. On **L21**: you're right that Cuccaro §4.1 reaches **2n−3** via an (n−1)-bit adder + final CNOT — I stopped at 2n−2 (the local cancellation) to keep the plain MAJ/UMA structure and said so in the docstring. Happy to adopt the 2n−3 construction if you'd prefer it.
- **Comparators**: applied the one-Toffoli saving — write the top carry straight into `out` (`CX(w;f) CCX(u,v;f)`), leaving `work[n-1]` untouched. `cmp_ge_constant` **and** `cmp_ge_register`/`cmp_gt_register` are now **2n−1** (Cuccaro's comparator count).
- Counts now live on each kernel docstring and the tests read that number; the doubling-law test asserts the affine slope (2n−2 doesn't double).

### Overflow (L324)
Fixed. `phase_add_constant` reduces `K mod 2^(t+1)` **as an integer** per bit before the rotation angle, so it stays in [0,2π) and is exact for any K, including K ≥ 2⁵³.

### Naming / shift (L380, L166, L377)
Renamed `cmp_ge_constant_qft` → **`cmp_ge_constant_qft_shift`** (+ `_shift_adj`); the docstring now leads with "leaves `x_reg` shifted to `(x−K) mod 2^n`" and contrasts it with `cmp_ge_constant` (x-preserving, out XOR-loaded, complement bits). The consumer that needs the shift is the alias sampler in #48.

### `cudaq.control` (L46 / test L30)
Right to distrust it — removed the "composes under `cudaq.control`" claim. On 0.15.1 direct control of several of these fails specialization and the wrapper fires on both branches; I'll file the upstream issue with a minimal reproducer and add a skip-gated superposed-control test that un-skips once it's fixed.

### Smaller
- **L162** `add_constant` is the straightforward CDKM-with-loaded-constant, not a placeholder; noted Häner (arXiv:1611.07995) as the ancilla-light follow-up on the kernel.
- **L307** `cmp_gt_register` kept as a named kernel for a discoverable API, with the docstring stating it's `cmp_ge_register` with the operands swapped + the strict X.
- **L17** module docstring trimmed to the conventions + arXiv links; per-gate prices moved onto the kernels.

Full arithmetic suite green (92).

wsttiger:
Follow-up on the `cudaq.control` thread (L27/L46): I bisected it.

`cudaq.control` of a device kernel that **calls a sub-kernel** fails on **0.15.1** with the specialization error, but is correct on **0.14.2** (before) and again on **0.16.0** (after) — a 0.15.x-only regression, already fixed upstream. Flat kernels (no sub-kernel call) control fine even on 0.15.1; the failing cases are exactly the ones that call sub-kernels (`add_constant`, `subtract_constant`, `cmp_gt_register`, the QFT wrappers).

Since the library targets 0.16, controlled composition genuinely works — so rather than file an already-fixed bug, I added a **version-gated superposed-control test** (d9bc49a): it controls a composed kernel (`add_constant_qft`) with the control in superposition and checks the result numerically, skipping below 0.16. A superposed control also catches a control that fires on *both* branches, so it doubles as the "is it really controlled?" check.

(Aside: on 0.15.1 I couldn't reproduce the silent "fires on both branches" you saw — here the composed case hard-fails specialization rather than mis-controlling. Either way it's gone in 0.16, so it's moot for the library.)

PR #48, opened 2026-09-08
Add alias sampling

wsttiger:
## Summary

Adds `AliasSamplingPrepare`, an implementation of the PREPARE-with-garbage construction described by Babbush et al. in [arXiv:1805.03662, Sec. III.D](https://arxiv.org/abs/1805.03662).

Given nonnegative weights `w_k`, the primitive prepares a state whose index-register measurement probabilities exactly match an integer-discretized alias table. This construction is a standard state-preparation component for qubitized quantum simulation.

The implementation builds on the arithmetic primitives introduced earlier in this PR stack:

- QROM loads the `(alias, keep)` pair associated with each bin.
- A CDKM comparator compares the keep threshold with a uniform reference register.
- Controlled swaps replace rejected bins with their aliases.
- A hand-written adjoint reverses the complete preparation.

> This PR is stacked on the reversible-arithmetic PR.

## Exact, bounded discretization

The input weights are discretized to `mu` bits using an integer version of Vose's alias-table algorithm.

The resulting quantum circuit realizes the discretized distribution exactly. Tests compare against exact rational probabilities rather than relying only on floating-point tolerances.

For `K` bins, the error for each probability is bounded by

`1.5 / (K * 2^mu)`.

The module docstring derives this bound, including why the residual-redistribution cursor cannot wrap around. Zero-weight entries and bins introduced through power-of-two padding receive exactly zero probability.

## Resource costs and register layout

The implementation exposes and tests the following resource contract:

- Toffoli count:

  `qrom.toffoli_count + 4 * (mu + 1)`

- Fredkin count:

  `num_index`

The resource tests validate these formulas against the compiled kernel using `cudaq.estimate_resources`. They also cross-check the QROM object's reported costs against the compiled QROM circuit.

The garbage-register layout is documented field by field. Its total width is

`num_garbage = num_index + 2 * mu + 4 + qrom.num_ladder`.

Because `qrom.num_ladder` depends on the selected QROM implementation, the garbage width automatically reflects the chosen variant.

The internal QROM accepts `variant=` and `block_size=` options. The default `"auto"` mode evaluates the available constructions and selects the lowest-cost option. Tests also force the `select_swap` variant to verify that both PREPARE and its adjoint are independent of the selected QROM implementation, including clean-ladder reuse during uncomputation.

## PREPARE-with-garbage contract

`AliasSamplingPrepare` guarantees the probability associated with each index value. It does **not** guarantee clean state preparation of the form

`sum_k sqrt(p_k) |k>|0>`.

Instead, the prepared state generally has the form

`sum_k sqrt(p_k) |k>|G_k>`,

where the conditional garbage states `|G_k>` need not be mutually orthogonal. Consequently, the reduced state of the index register is not necessarily `diag(p_k)`.

This is sufficient for the qubitization identity

`<0| PREP^dag SELECT PREP |0> = sum_k p_k U_k`.

Because `SELECT` is diagonal in the index basis, index orthogonality eliminates cross terms regardless of overlaps between the garbage states. The class documentation states this contract explicitly so consumers do not assume clean amplitude preparation or a diagonal reduced index state.

## Validation

The test suite includes:

- Exact index-marginal comparisons against independently generated alias tables.
- Brute-force enumeration of every `(bin, reference)` branch and comparison with circuit results.
- PREPARE-followed-by-adjoint identity checks.
- Resource-identity checks for both the default and forced QROM variants.
- Exact checks that zero-weight and padded bins remain zero.
- An independent adversarial sweep over approximately 4,000 weight tables, including:
  - Dynamic ranges from `exp(-18)` through `exp(4)`.
  - Irrational weights.
  - Tables containing many zeros.
  - Precision values through `mu = 8`.

The adversarial sweep found no discretization-bound violations, and padded bins remained exactly zero.

At this branch tip:

- 26 tests are added by this PR.
- The complete suite reports 560 passed and 3 skipped.

## Deferred work

The following extensions are deliberately left for future changes:

- Sub-bit-precision variants.
- Measurement-assisted QROM uncomputation.

Both can be introduced behind the existing QROM interface when a consumer requires the associated resource tradeoff.

### DIFF
--- a/python/cudaq_algorithms/primitives/_alias_sampling.py
+++ b/python/cudaq_algorithms/primitives/_alias_sampling.py
@@ -1,0 +1,385 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""Coherent alias sampling: PREPARE-with-garbage for weighted indices.
+
+``AliasSamplingPrepare(weights, mu)`` mints the Babbush-style
+(`arXiv:1805.03662`, Sec. III.D) state preparation whose *index-register
+marginal* is the normalized weight distribution ``w / lambda`` up to a
+``mu``-bit discretization: classical Vose preprocessing builds integer
+keep/alias tables, and the circuit is
+
+1. uniform superposition over the ``K = 2^num_index`` bins (``weights``
+   is zero-padded to a power of two; padded bins carry exactly zero
+   probability),
+2. a :class:`~cudaq_algorithms.primitives.QROM` lookup of
+   ``(alias_k, keep_k)`` at bin ``k`` (``variant="auto"``: the lookup is
+   priced across the available QROM variants and the cheapest
+   construction is minted — its clean-ancilla ladder lives inside the
+   garbage register, so ``num_garbage`` follows the chosen variant),
+3. a comparator of ``keep_k`` against a uniform ``mu``-bit reference
+   register (built from the CDKM register adders of ``_arithmetic``:
+   subtract on a ``mu+1``-bit extension, copy the borrow, add back), and
+4. a controlled swap of the bin index with the alias when the reference
+   is not below ``keep_k``.
+
+PREPARE with garbage — read this before consuming
+-------------------------------------------------
+
+The output is **not** ``sum_k sqrt(w_k / lambda) |k> |0...0>``: the
+alias/keep/reference/flag registers stay *entangled* with the index (only
+the amplitude *magnitudes* marginalize to ``w / lambda``). That is
+exactly the coherent-alias-sampling contract: qubitization-style
+consumers tolerate the garbage only because every use is the symmetric
+sandwich ``PREPARE ... PREPARE^dagger`` — the garbage registers are
+uncomputed by the hand-written ``adjoint_kernel()`` applied to the same
+registers, never discarded, never reflected over while dirty, and never
+consumed by any other circuit element in between. Anything that needs a
+clean ``sum sqrt(p_k) |k>`` (e.g. amplitude arithmetic on the index
+alone) must use a different preparation.
+
+Discretization: the integer tables represent bin probabilities exactly as
+``W_k / (K 2^mu)`` (integer Vose is exact, and full bins are stored
+self-aliased so the ``keep = 2^mu`` edge loses nothing); the only error
+is rounding ``w_k / lambda`` to the integers ``W_k``, bounded per bin by
+``1.5 / (K 2^mu)`` (half-ulp rounding plus at most one unit of residual
+redistribution) — exposed as ``discretization_bound`` and derived, never
+assumed, in the tests.
+
+Kernel signatures (little-endian; ``docs/conventions.md``): both
+``kernel()`` and ``adjoint_kernel()`` are ``(index: qview, garbage:
+qview)`` with ``index`` of width ``num_index`` and ``garbage`` of width
+``num_garbage`` laid out as ``[alias(num_index) | keep(mu) | keep_pad |
+ref(mu) | ref_pad | flag | ladder(qrom.num_ladder) | carry]`` — i.e.
+``num_garbage = num_index + 2 mu + 3 + qrom.num_ladder + 1``, which is
+``2 num_index + 2 mu + 4`` whenever the priced lookup is the plain
+``"select"`` walk (``num_ladder = num_index``; always the case for the
+small tables where select wins ``"auto"``'s pricing). Both registers
+must be |0...0> on entry to ``kernel()``; ``adjoint_kernel()`` is the
+literal gate-reversal (no ``cudaq.adjoint``, cuda-quantum#4897/#4898) and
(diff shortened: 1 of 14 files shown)


python/cudaq_algorithms/primitives/_alias_sampling.py
```diff
@@ -0,0 +1,385 @@
...
+        size through ``qrom.num_ladder`` (see ``num_garbage``).
+    block_size
+        Keyword-only; forwarded verbatim to the :class:`QROM`
+        constructor (only meaningful with ``variant="select_swap"``;
+        ``None`` lets the QROM pick). Also shapes ``qrom.num_ladder``
+        and hence ``num_garbage``. Invalid combinations raise the
+        QROM's own ``ValueError``.
+
+    Garbage caveat: the per-bin garbage vectors ``|g_k>`` are NOT
+    mutually orthogonal (measured off-diagonal overlaps reach ~0.19), so
```
smcardleQ: Remove "(measured off-diagonal overlaps reach ~0.19)" from docstring, as it is instance specific without details.

python/cudaq_algorithms/primitives/_alias_sampling.py
```diff
@@ -0,0 +1,385 @@
...
+        self._probabilities = np.asarray(probabilities)
+        self._table_probabilities = np.asarray(table)
+
+        # QROM data: alias in the low num_index bits, keep above it.
+        # variant/block_size pass through verbatim (default "auto"
+        # prices the constructions); the chosen variant's ladder width
+        # shapes the garbage layout below. QROM validation errors
+        # propagate as-is.
+        self._qrom = QROM(
+            [alias[k] | (keep[k] << num_index) for k in range(num_bins)],
```
smcardleQ, line 228: Add a comment on the packing convention here, as the one-liner made it non-obvious for me.

python/cudaq_algorithms/primitives/_alias_sampling.py
```diff
@@ -0,0 +1,385 @@
...
+        # Garbage layout offsets (see the module docstring); everything
+        # unpacked into scalar locals (no tuple/self capture in kernels).
+        m = self._num_index
+        mu = self._mu
+        num_ladder = self._qrom.num_ladder
+        k0 = m  # keep (keep_pad at k0 + mu)
+        r0 = m + mu + 1  # ref (ref_pad at r0 + mu)
+        flag = m + 2 * mu + 2
+        l0 = m + 2 * mu + 3
+        c0 = l0 + num_ladder  # carry
```
smcardleQ: k0, r0, l0, c0 are not sufficiently informative as variable names.

python/cudaq_algorithms/primitives/_alias_sampling.py
```diff
@@ -0,0 +1,385 @@
...
+            x(garbage[flag])
+            cx(garbage[r0 + mu], garbage[flag])
+            add_register(garbage[k0:k0 + mu + 1], garbage[r0:r0 + mu + 1],
+                         garbage[c0:c0 + 1])
+            # The QROM lookup uncomputes itself on the clean-ladder
+            # sector: the ladder is |0> here (restored by the compute
+            # pass, untouched since), so re-applying the lookup XORs the
+            # table back out of garbage[0 : m + mu] — the documented
+            # self-inverse contract of every QROM variant.
+            qrom_kernel(index, garbage[l0:l0 + num_ladder], garbage[0:m + mu])
```
smcardleQ: Does the inverse QROM assume the same ladder ancilla qubits as the forward QROM? The ladder ancillas will be left clean (in |0>) after the forward QROM - hence they might be used by some other subroutine in the algorithm. We may need to allocate new ladder ancillas to help the inverse QROM.
This isn't a deal-breaker; one could instead reserve the ladder ancillas, and allocate other ancillas in the meantime. But potentially something for us to be aware of.

smcardleQ: Current implementation is fine. However, in future we may want to explicitly make this a "qrom_inverse" call. The reason being is that once we allow measurement-based uncomputation, uncomputing the QROM can be significantly cheaper than forward pass computing it, making the inverse alias sampling much cheaper than the forward. Leave a "To Do" ?

smcardleQ (approved):
Minor comments on coherent alias sampling, otherwise looks good to merge. I have not checked the arithmetic primitives yet.

python/cudaq_algorithms/primitives/_alias_sampling.py
wsttiger: Yes — the adjoint reuses the same ladder ancillas (`garbage[ladder_off : ladder_off + num_ladder]`), which the forward pass leaves clean in |0⟩, so the self-inverse lookup is valid. You're right that this reserves those qubits across the PREPARE/PREPARE† sandwich rather than freeing them for other subroutines in between; since coherent alias sampling is only ever used as that symmetric sandwich, holding them is intentional. If a consumer needs them freed mid-sandwich we'd allocate separate ancillas for the inverse — noted. (Also left a TODO at this site for the related `qrom_inverse`/measurement-based-uncompute point in your other comment.)

PR #50, opened 2026-09-14
Require cudaq >= 0.16.0

wsttiger:
## Description

CUDA-Q 0.16.0 is released; this makes it the minimum supported
version. Four coordinated changes:

- `pyproject.toml`: the wheel now declares `cudaq >= 0.16.0, < 0.17`.
- CI pip lanes install the same range (kept in lockstep by design —
  PRs test exactly the claim the wheel ships).
- The Custom-mode wheel version placeholder moves 0.15.99 → 0.16.99 to
  stay above the released version line.

## Validation

- Full suite run against the released `cudaq==0.16.0` wheel ahead of
  this PR: 472 passed / 3 skipped on the primitives line (a superset
  of every test on main). v0.1.0-era functionality requires no code
  changes on 0.16.
- The one 0.16-sensitive construction (1-bit-address QROM lookups in
  the unary-iteration feature) ships with its own protection — the
  dispatch-pruning commit already on #40 — so this PR and #40 can
  merge in either order. Open PRs adopt the new range only when they
  rebase, at which point they inherit that commit too.
- This PR's own CI validates the switch: the pip lanes install 0.16.0
  and run the full suite against it.
- The `.cudaq_version` pin bump is deliberately split into a follow-up
  PR: its first firing of the from-source lanes surfaced a
  build-layout change in the 0.16 source tree (built install no longer
  exposes the python package on the expected prefix path) — the lanes
  doing exactly the job #42's review designed them for. The pin moves
  once the build plumbing adapts.

## Notes

- 0.16.0 ships without the fix for cuda-quantum#5280 (verified in the
  release tag); with the pruning commit merged, nothing in this
  library triggers it.
- 0.16 lifts the control-variant restriction on kernels calling
  kernels; follow-up modernization (composed select_swap, composition
  contract) is staged on `features/primitives_016_refactor` and lands
  separately after this.


anjbur (changes requested):
We'll also need to bump the `0.15.99` version used in `get-cudaq-wheels/action.yaml` up to `0.16.99`. Otherwise, the CI will break for Custom wheel builds.

PR #51, opened 2026-09-14
Fix the source-lane shell dispatch; pin cudaq to the 0.16.0 release

wsttiger:
## Description

Follow-up to #50, carrying the `.cudaq_version` bump that was split out
of it — together with the fix for what its first run caught.

**The bug** (pre-existing since #42, first exercised by #50's pin
change): the `Run tests` step in `lib_algorithms.yaml` branches on
`if [[ "$cudaq_source" == "source" ]]`, but the step has no `shell:`
and resolves to `sh` (dash) inside the CI container — where `[[` does
not exist. The failed condition silently took the pip branch
(`--pip-cudaq`, which deliberately puts no source install on
`PYTHONPATH` and whose pip install is gated to pip lanes), producing
`ModuleNotFoundError: No module named 'cudaq'` on all four source
lanes. The masking was perfect on ordinary PRs: for pip lanes the
broken conditional happens to take the correct branch. Evidence: the
run log shows `shell: sh -e {0}` and `[[: not found` one line before
the import error; the CUDA-Q 0.16.0 build itself succeeded and
installed `cudaq/__init__.py` exactly where `--cudaq-prefix` looks.

**The fix**: a workflow-level `defaults: run: shell: bash` on
`lib_algorithms.yaml` (with a comment on the dash hazard), covering
this step and the same latent issue in the Configure step.

**The pin**: `.cudaq_version` moves to the 0.16.0 release tag commit
(`51671baa`), so this PR's own CI fires the from-source lanes through
the repaired conditional — building the released source (the build
from #50's run is already cached under the matching key) and running
the suite against it. That run is the validation.

---

**Update (commit 638e76d):** the pin change also surfaced a second
latent bug — both reusable build workflows keyed their concurrency
group on `github.ref` alone, so when `pr_workflow` invokes
`lib_algorithms` twice in one run (pip + source lanes) the second
invocation cancelled the first (visible as the Algorithms lanes
repeatedly cancelling). Fixed by scoping the group with the
`cudaq_source` / `cudaq_wheels` input. Two-for-one: the source lanes
were designed to catch exactly this class of "only shows up when the
pin moves" breakage.


anjbur (approved):
Looks good to me as long as PR #50 is merged first, so that the CI can verify that the version bumps don't introduce any breaks in the Custom wheel flow.

